PR #58, opened 2023-10-23
Add link to stable docs

ranocha:


JoshuaLampert (changes requested):
Thanks! This should probably also go [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/main/docs/src/index.md).

ranocha:
Good catch! I just added the badge there, too

JoshuaLampert:
I just saw that the link to the docs in the README [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/main/README.md?plain=1#L14) refers to the dev branch. Maybe update this accordingly, too?

PR #64, opened 2023-11-07
Interpolate x value for plotting

JoshuaLampert:
I was hoping to improve the plots by not using the constant interpolation at a position `x`, but using higher order interpolation, in particular I was hoping to have a smaller phase shift in some plots, where we plot the solution at certain `x` values that are not on the grid. Unfortunately, the plots did not really improve, also not for cubic interpolation.


ranocha:
Nice idea. You could check how it looks like in a scatter plot of the numerical data overlayed on the reference data. If you don't see an improvement of the phase shift there, there is no chance to get better results using intrpolations.

PR #89, opened 2024-02-10
Fix formatter version

JoshuaLampert:
See https://github.com/trixi-framework/Trixi.jl/pull/1842.

### DIFF
--- a/.github/workflows/Format-check.yml
+++ b/.github/workflows/Format-check.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install JuliaFormatter and format
         run: |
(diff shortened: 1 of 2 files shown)


ranocha:
> Let's wait for [domluna/JuliaFormatter.jl#807](https://github.com/domluna/JuliaFormatter.jl/pull/807).

I agree. Alternatively, we could also fix the version of JuliaFormatter to an older one without the change of the formatting behavior.

JoshuaLampert:
I fixed the version of the JuliaFormatter like we do in Trixi.jl, see https://github.com/trixi-framework/Trixi.jl/pull/1843.

ranocha (approved):
Thanks

PR #91, opened 2024-02-23
Fix initial condition from Dingemans experiment for BBM-BBM equations

JoshuaLampert:
The results from the BBM-BBM equations for the experimental setup was missing a vertical translation of the water surface and the bathymetry. This is fixed here and the notation about the bathymetry $b$ and still water depth $D$ is used consistently.
We have $D = \eta_0 - b$ for the still water surface $\eta_0$. The still water surface can be arbitrary in the case of the Svärd-Kalisch equations, but needs to be 0 for the BBM-BBM equations (otherwise there would be an additional term in the equations).


src/equations/bbm_bbm_1d.jl
```diff
@@ -148,15 +151,15 @@ end
 @inline function prim2cons(q, equations::BBMBBMEquations1D)
     eta, v = q
 
-    h = eta + equations.D
+    h = eta - equations.eta0 + equations.D
```
JoshuaLampert, line 154: Although `equations.eta0` needs to be 0, I use it in the conversions such that it can be generalized more easily in the future if required and such that it matches the general definitions. IMO, this is a bit more understandable, but it might be less understandable because it is always 0.

src/equations/svaerd_kalisch_1d.jl
```diff
@@ -327,7 +326,7 @@ number of nodes as length of the second dimension.
     eta = view(q, 1, :)
     v = view(q, 2, :)
     D = view(q, 3, :)
-    beta_hat = equations.beta * (eta .+ D) .^ 3
```
JoshuaLampert, line 330: This was actually a bug. It should have been `D` already before. That's also why there are some changes in the change of the modified entropy in the test values.

src/equations/bbm_bbm_1d.jl
ranocha: 👍

src/equations/svaerd_kalisch_1d.jl
ranocha: 👍

ranocha (approved):
Thanks!

PR #92, opened 2024-02-25
Fix initial condition for manufactured solution (Svärd-Kalisch equations)

JoshuaLampert:
In #91 I forgot to adjust the initial condition for the manufactured solution for the Svärd-Kalisch equations. It only worked for $\eta_0 = 0$. Now this is fixed and should work for general $\eta_0$.


ranocha (approved):
Thanks!

PR #96, opened 2024-03-11
Add test using ExplicitImports.jl

JoshuaLampert:
With https://github.com/ericphanson/ExplicitImports.jl there exists a tool to avoid implicit imports and avoid the existence of unused imports. I updated the imports accordingly and added a regression test. I put the test next to the Aqua tests since this is kind of similar.

### DIFF
--- a/test/test_aqua.jl
+++ b/test/test_aqua.jl
@@ -1,7 +1,7 @@
 module TestAqua
 
 using Aqua
-using ExplicitImports
+using ExplicitImports: check_no_implicit_imports, check_no_stale_explicit_imports
 using Test
 using DispersiveShallowWater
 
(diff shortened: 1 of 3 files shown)


test/test_aqua.jl
```diff
@@ -1,12 +1,15 @@
 module TestAqua
 
 using Aqua
+using ExplicitImports
```
ranocha, line 4:
```suggestion
using ExplicitImports: check_no_implicit_imports, check_no_stale_explicit_imports
```

ranocha (approved):
Thanks! I just have a minor suggestion.

test/test_aqua.jl
JoshuaLampert: Speaking of explicit imports... :sweat_smile:

ranocha (approved):

PR #105, opened 2024-05-18
Add test for `BBMBBMEquations1D` with upwind operators

JoshuaLampert:
We didn't test the constant topography `BBMBBMEquations1D` with an upwind solver before.

### DIFF
--- a/test/test_bbm_bbm_1d.jl
+++ b/test/test_bbm_bbm_1d.jl
@@ -35,22 +35,28 @@ EXAMPLES_DIR = joinpath(examples_dir(), "bbm_bbm_1d")
         using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
         @test_trixi_include(joinpath(EXAMPLES_DIR, "bbm_bbm_1d_dg.jl"),
                             tspan=(0.0, 1.0),
-                            D1=upwind_operators(periodic_derivative_operator; derivative_order = 1,
-                              accuracy_order = 4, xmin = -35.0, xmax = 35.0,
-                              N = 512),
-                            D_pl=upwind_operators(periodic_derivative_operator; derivative_order = 1,
-                              accuracy_order = 4, xmin = -35.0, xmax = 35.0,
-                              N = 512).plus,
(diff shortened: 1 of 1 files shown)


test/test_bbm_bbm_1d.jl
```diff
@@ -31,6 +31,33 @@ EXAMPLES_DIR = joinpath(examples_dir(), "bbm_bbm_1d")
...
+                            D_pl=upwind_operators(periodic_derivative_operator;
+                                                  derivative_order = 1,
+                                                  accuracy_order = 4, xmin = -35.0,
+                                                  xmax = 35.0,
+                                                  N = 512).plus,
+                            D_min=upwind_operators(periodic_derivative_operator;
+                                                   derivative_order = 1,
+                                                   accuracy_order = 4, xmin = -35.0,
+                                                   xmax = 35.0,
+                                                   N = 512).minus,
```
JoshuaLampert, line 52: This is not very nice of course. I would like to define `D1` before the `@test_trixi_include` and reuse it in the macro call, but this is not possible due to `D1` then not being defined. Of course, I could add a new elixir with the upwind operators, but I thought it would be too similar to two of the existing ones, so it's more appropriate to just add it to the tests. Do you know of a nicer way to just change the solver in the tests, @ranocha?

JoshuaLampert: I just pushed another alternative in [05d607f](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/105/commits/05d607f34830417d0425bc030fbedb4f0717bb4f), where I explicitly recreate the `semi` and the `ode` and test the errors outside of `@test_trixi_include`. This avoids passing the ugly SBP operators to `@test_trixi_include`, but has also some code duplication because the errors comparisons are done explicitly. However, this approach is more generally applicable (e.g. if we would like to exchange the whole solver object by something like `Solver(D1, D2)`, where, e.g., `D2 = sparse(D1)^2` this would not be really doable with the first approach).

ranocha: I think that's fine - I don't really have a better suggestion for now

ranocha (approved):

PR #106, opened 2024-05-19
Refactor `create_cache`

JoshuaLampert:
This passes the `boundary_conditions` to `create_cache` and dispatches on them instead of looking at the type of the SBP operators to infer whether the problem has periodic or reflecting boundary conditions, e.g. previously it was possible to use `boundary_conditions_reflecting` and pass periodic operators to the `Solver`, which meant that the `create_cache` function also created the cache for the periodic problem instead of for reflecting. This was a bit messy. This way it should work more as expected and is also easier to extend.
On the fly, I put `tmp1`(needed by the `RelaxationCallback`) into the `intitial_cache` instead of always creating it in `create_cache`, which is weird because `tmp1` is sometimes not needed for the discretization itself.
Finally, I fixed the upwind discretization for the reflecting boundary conditions for the `BBMBBMEquations1D`, which was broken before. I also added a test in a similar manner as in #105.

### DIFF
--- a/src/equations/bbm_bbm_1d.jl
+++ b/src/equations/bbm_bbm_1d.jl
@@ -154,15 +154,15 @@ function create_cache(mesh, equations::BBMBBMEquations1D,
 
     # homogeneous Neumann boundary conditions
     if solver.D1 isa DerivativeOperator ||
-        solver.D1 isa UniformCoupledOperator
+       solver.D1 isa UniformCoupledOperator
         D1_b = BandedMatrix(solver.D1)
         invImD2n = inv(I + 1 / 6 * D^2 * inv(M) * D1_b' * PdM * D1_b)
-     elseif solver.D1 isa UpwindOperators
+    elseif solver.D1 isa UpwindOperators
         D1plus_b = BandedMatrix(solver.D1.plus)
         invImD2n = inv(I + 1 / 6 * D^2 * inv(M) * D1plus_b' * PdM * D1plus_b)
-     else
-         @error "unknown type of first-derivative operator: $(typeof(solver.D1))"
-     end
+    else
+        @error "unknown type of first-derivative operator: $(typeof(solver.D1))"
+    end
     return (invImD2d = invImD2d, invImD2n = invImD2n)
 end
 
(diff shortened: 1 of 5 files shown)


src/equations/bbm_bbm_1d.jl
```diff
@@ -127,39 +127,43 @@ function source_terms_manufactured_reflecting(q, x, t, equations::BBMBBMEquation
...
     D = equations.D
-    if solver.D1 isa PeriodicDerivativeOperator ||
-       solver.D1 isa UniformPeriodicCoupledOperator ||
-       solver.D1 isa PeriodicUpwindOperators
-        invImD2 = inv(I - 1 / 6 * D^2 * Matrix(solver.D2))
-        return (invImD2 = invImD2, tmp1 = tmp1)
-    elseif solver.D1 isa DerivativeOperator ||
-           solver.D1 isa UniformCoupledOperator ||
-           solver.D1 isa UpwindOperators
+    invImD2 = inv(I - 1 / 6 * D^2 * Matrix(solver.D2))
```
ranocha, line 135: In the long run, one should consider allowing more flexibility here. For example, FD methods should use sparse matrices and just a factorization instead of the inverse of a dense matrix. But that's not critical right now.

src/equations/bbm_bbm_1d.jl
```diff
@@ -127,39 +127,43 @@ function source_terms_manufactured_reflecting(q, x, t, equations::BBMBBMEquation
...
+function create_cache(mesh, equations::BBMBBMEquations1D,
+                      solver, initial_condition,
+                      ::BoundaryConditionReflecting,
+                      RealT, uEltype)
+    D = equations.D
+    N = nnodes(mesh)
+    M = mass_matrix(solver.D1)
+    Pd = BandedMatrix((-1 => fill(one(real(mesh)), N - 2),), (N, N - 2))
+    D2d = (sparse(solver.D2) * Pd)[2:(end - 1), :]
+    # homogeneous Dirichtlet boundary conditions
```
ranocha, line 148:
```suggestion
    # homogeneous Dirichlet boundary conditions
```

ranocha (dismissed):
Thanks! I just have some minor comments that can either be resolved quickly or maybe postponed to later development (issues). Feel free to merge :+1:

src/equations/bbm_bbm_1d.jl
JoshuaLampert: Yeah, you are definitely right. IIRC, when I tried that some time ago, I had some problems making that properly work, but I don't remember the exact issue. Back then I focused on something that just works and is easy to implement andunderstand . But of course this is not the most performant way. I created #107 for that.

Issue #107, opened 2024-05-20
Improve efficiency


JoshuaLampert:
We store the whole unfactorized dense inverses of the relevant matrices. It would be more efficient to store them in a factorized way and/or use sparse matrices, see also https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/106#discussion_r1606998619.

TODO:

- [x] `BBMBBMEquations1D`: #108
- [x] `BBMBBMVariableEquations1D`: #112
- [ ] `SvaerdKalischEquations1D`: #114

JoshuaLampert:
For the Svärd-Kalisch equations, it's a bit more delicate because the elliptic operators changes over time, which currently leads to massive allocations:
https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/3adee6615091d6739f35bcae9970e7d8505a7b7c/src/equations/svaerd_kalisch_1d.jl#L259-L262

The `SummaryCallback` for `examples/svaerd_kalisch_1d/svaerd_kalisch_1d_dingemans.jl` currently gives
```julia
 ──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations      
                             ───────────────────────   ────────────────────────
      Tot / % measured:           1.91s /  99.3%           3.61GiB /  99.7%    

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 5.16k    1.88s   99.2%   364μs   3.59GiB   99.7%   730KiB
   dv elliptic        5.16k    1.10s   58.2%   213μs   2.77GiB   76.9%   563KiB
   dv hyperbolic      5.16k    719ms   38.0%   139μs    590MiB   16.0%   117KiB
   deta hyperbolic    5.16k   53.3ms    2.8%  10.3μs    250MiB    6.8%  49.5KiB
   ~rhs!~             5.16k   4.87ms    0.3%   944ns    810KiB    0.0%     161B
   source terms       5.16k    265μs    0.0%  51.3ns     0.00B    0.0%    0.00B
 analyze solution        86   14.3ms    0.8%   167μs   10.0MiB    0.3%   119KiB
 ──────────────────────────────────────────────────────────────────────────────
```

Do you see a possibility to reduce the allocations @ranocha?

ranocha:
I'm using the following approach in another project:

```julia
system_matrix = Symmetric(...)
[...]

PR #108, opened 2024-05-21
Use factorized matrices in `BBMBBMEquations1D`

JoshuaLampert:
A first try to tackle #107 for the `BBMBBMEquations1D`. The `SummaryCallback` gives the following after an initial run to exclude compilation time:

Current `main`:
```julia
──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations
                             ───────────────────────   ────────────────────────
      Tot / % measured:           1.38s /  98.7%            762MiB /  99.9%

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 16.8k    1.33s   97.8%  79.5μs    745MiB   98.0%  45.5KiB
   deta elliptic      16.8k    535ms   39.2%  31.9μs   67.5MiB    8.9%  4.12KiB
   dv elliptic        16.8k    517ms   38.0%  30.9μs   67.5MiB    8.9%  4.12KiB
   dv hyperbolic      16.8k    141ms   10.3%  8.42μs    338MiB   44.4%  20.6KiB
   deta hyperbolic    16.8k    128ms    9.4%  7.66μs    270MiB   35.5%  16.5KiB
   ~rhs!~             16.8k   10.7ms    0.8%   639ns   2.56MiB    0.3%     160B
   source terms       16.8k    640μs    0.0%  38.2ns     0.00B    0.0%    0.00B
 analyze solution       280   30.2ms    2.2%   108μs   15.4MiB    2.0%  56.5KiB
 ──────────────────────────────────────────────────────────────────────────────
```
This branch
```julia
  ──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations
                             ───────────────────────   ────────────────────────
      Tot / % measured:           680ms /  98.0%           1.66GiB /  99.9%

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 16.8k    635ms   95.2%  37.9μs   1.64GiB   99.1%   103KiB
   dv elliptic        16.8k    175ms   26.2%  10.4μs    534MiB   31.5%  32.6KiB
   deta elliptic      16.8k    174ms   26.1%  10.4μs    534MiB   31.5%  32.6KiB
   dv hyperbolic      16.8k    147ms   22.0%  8.77μs    338MiB   19.9%  20.6KiB
   deta hyperbolic    16.8k    127ms   19.0%  7.55μs    270MiB   15.9%  16.5KiB
   ~rhs!~             16.8k   11.6ms    1.7%   690ns   2.64MiB    0.2%     165B
   source terms       16.8k    579μs    0.1%  34.6ns     0.00B    0.0%    0.00B
 analyze solution       280   32.1ms    4.8%   115μs   15.4MiB    0.9%  56.5KiB
 ──────────────────────────────────────────────────────────────────────────────
```
So the new version is almost twice as fast as the old one, which is good. But I feel like we can avoid the many allocations in the new version. I tried using `ldiv!(invImD2, deta)`, but this gave the following error message:
```julia
ERROR: LoadError: MethodError: no method matching ldiv!(::SparseArrays.CHOLMOD.Factor{Float64, Int64}, ::SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true})
```
Any ideas to improve this further, @ranocha? After we settled on an efficient way, I can apply it to the other equations, too.


JoshuaLampert:
Using `lu` instead of `factorize` gives:

 ```julia
──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations
                             ───────────────────────   ────────────────────────
      Tot / % measured:           586ms /  96.4%            904MiB /  99.1%

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 16.8k    536ms   94.8%  31.9μs    880MiB   98.3%  53.8KiB
   deta elliptic      16.8k    137ms   24.3%  8.18μs    135MiB   15.1%  8.25KiB
   dv elliptic        16.8k    135ms   23.9%  8.07μs    135MiB   15.1%  8.25KiB
   dv hyperbolic      16.8k    129ms   22.9%  7.72μs    338MiB   37.7%  20.6KiB
   deta hyperbolic    16.8k    126ms   22.3%  7.52μs    270MiB   30.2%  16.5KiB
   ~rhs!~             16.8k   7.18ms    1.3%   428ns   2.56MiB    0.3%     160B
   source terms       16.8k    538μs    0.1%  32.1ns     0.00B    0.0%    0.00B
 analyze solution       280   29.6ms    5.2%   106μs   15.4MiB    1.7%  56.5KiB
 ──────────────────────────────────────────────────────────────────────────────
```
Again, `ldiv!` doesn't work. This time because of `ERROR: LoadError: ArgumentError: in and output vectors must have unit strides`.

**Review by ranocha — COMMENTED** [review]
There are several cases we could consider for optimal performance:

- For Fourier (and periodic FD) methods, SummationByPartsOperators.jl implements rational functions of operators efficiently using FFT (with `mul!` or `*` interface), e.g., 
```julia
julia> using SummationByPartsOperators, LinearAlgebra

julia> D = fourier_derivative_operator(xmin = 0.0, xmax = 1.0, N = 8)
Periodic 1st derivative Fourier operator {T=Float64} 
on a grid in [0.0, 1.0] using 8 nodes and 5 modes

julia> inv(I - 2 * D^2)
Rational Fourier operator with coefficients
(1.0,)
and
(1.0, -0.0, -2.0)
of the operator:
Periodic 1st derivative Fourier operator {T=Float64} 
on a grid in [0.0, 1.0] using 8 nodes and 5 modes

julia> inv(I - 2 * D^2) |> Matrix
8×8 Matrix{Float64}:
 0.129366  0.126864  0.12431   0.122939  0.12241   0.122939  0.12431   0.126864
 0.126864  0.129366  0.126864  0.12431   0.122939  0.12241   0.122939  0.12431
 0.12431   0.126864  0.129366  0.126864  0.12431   0.122939  0.12241   0.122939
 0.122939  0.12431   0.126864  0.129366  0.126864  0.12431   0.122939  0.12241
 0.12241   0.122939  0.12431   0.126864  0.129366  0.126864  0.12431   0.122939
 0.122939  0.12241   0.122939  0.12431   0.126864  0.129366  0.126864  0.12431
 0.12431   0.122939  0.12241   0.122939  0.12431   0.126864  0.129366  0.126864
 0.126864  0.12431   0.122939  0.12241   0.122939  0.12431   0.126864  0.129366

julia> D = periodic_derivative_operator(derivative_order = 1, accuracy_order = 2,
                                        xmin = 0.0, xmax = 1.0, N = 8)
Periodic first-derivative operator of order 2 on a grid in [0.0, 1.0] using 8 nodes, 
stencils with 1 nodes to the left, 1 nodes to the right, and coefficients of Fornberg (1998) 
  Calculation of Weights in Finite Difference Formulas. 
  SIAM Rev. 40.3, pp. 685-691.

julia> inv(I - 2 * D^2)
Rational periodic operator with coefficients
(1.0,)
and
(1.0, 0.0, -2.0)
of the operator:
Periodic first-derivative operator of order 2 on a grid in [0.0, 1.0] using 8 nodes, 
stencils with 1 nodes to the left, 1 nodes to the right, and coefficients of Fornberg (1998) 
  Calculation of Weights in Finite Difference Formulas. 
  SIAM Rev. 40.3, pp. 685-691.

julia> inv(I - 2 * D^2) |> Matrix
8×8 Matrix{Float64}:
  0.25963       8.67362e-19   0.248062     …  -8.67362e-19   0.248062      8.67362e-19
  9.19976e-19   0.25963       9.19976e-19      0.244246     -9.19976e-19   0.248062
  0.248062      8.67362e-19   0.25963         -8.67362e-19   0.244246     -8.67362e-19
 -9.19976e-19   0.248062      9.19976e-19      0.248062     -9.19976e-19   0.244246
  0.244246     -8.67362e-19   0.248062         8.67362e-19   0.248062     -8.67362e-19
 -9.19976e-19   0.244246     -9.19976e-19  …   0.25963       9.19976e-19   0.248062
  0.248062     -8.67362e-19   0.244246         8.67362e-19   0.25963       8.67362e-19
  9.19976e-19   0.248062     -9.19976e-19      0.248062      9.19976e-19   0.25963
```
- Since we know that the system matrix is diagonal, we should wrap it accordingly as `Symmetric(...)` and use a `cholesky` factorization. The wrapping may be required since it may not be perfectly symmetric for upwind operators. Then, we can use the 3-argument version of `ldiv!(result, factorization, input)`.

JoshuaLampert:
Using your first approach (i.e. `invImD2 = inv(I - 1 / 6 * D^2 * solver.D2)`) gives `ERROR: LoadError: ArgumentError: FFTW plan applied to wrong-strides array` while `deta[:]=invImD2 * deta`.

JoshuaLampert:
Trying your second approach gives me `ERROR: LoadError: matrix does not have contiguous columns` while `ldiv!(deta, invImD2, deta)` (or `ldiv!(invImD2, deta)`)

JoshuaLampert:
Yes, I do. Just so that I understand: The problem is that `u_ode`, i.e. the data OrdinaryDiffEq.jl is working with, is just a vector and not the matrix of size `nvariables(equations), nnodes(mesh)`? So we could also consider just using the matrix as the data for OrdinaryDiffEq.jl and would also spare us some `wrap_array`s? IIRC, in Trixi.jl we need to store the data as a vector to be able to `resize!` it for AMR, right? We don't need to do this here. What do you think? Or am I misunderstanding something?

JoshuaLampert:
I tried your suggestion to use temporary arrays in [885c329](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/108/commits/885c329f1eb346f01962f6909a6d37f269c64833). This works with no allocations, but is much slower. Did you do something wrong? (Ignore the reflecting boundary conditions for now. The result below is for `examples/bbm_bbm_1d/bbm_bbm_1d_basic.jl`, so periodic bc.

```julia
 ──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations      
                             ───────────────────────   ────────────────────────
      Tot / % measured:           3.62s /  99.5%            627MiB /  99.8%    

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 16.8k    3.56s   99.0%   212μs    610MiB   97.5%  37.3KiB
   deta elliptic      16.8k    1.66s   46.1%  98.9μs     0.00B    0.0%    0.00B
   dv elliptic        16.8k    1.62s   45.0%  96.5μs     0.00B    0.0%    0.00B
   dv hyperbolic      16.8k    133ms    3.7%  7.92μs    338MiB   54.0%  20.6KiB
   deta hyperbolic    16.8k    130ms    3.6%  7.75μs    270MiB   43.2%  16.5KiB
   ~rhs!~             16.8k   23.3ms    0.6%  1.39μs   2.56MiB    0.4%     160B
   source terms       16.8k    709μs    0.0%  42.3ns     0.00B    0.0%    0.00B
 analyze solution       280   36.1ms    1.0%   129μs   15.4MiB    2.5%  56.5KiB
 ──────────────────────────────────────────────────────────────────────────────
```

ranocha:
> Yes, I do. Just so that I understand: The problem is that `u_ode`, i.e. the data OrdinaryDiffEq.jl is working with, is just a vector and not the matrix of size `nvariables(equations), nnodes(mesh)`? So we could also consider just using the matrix as the data for OrdinaryDiffEq.jl and would also spare us some `wrap_array`s? IIRC, in Trixi.jl we need to store the data as a vector to be able to `resize!` it for AMR, right? We don't need to do this here. What do you think? Or am I misunderstanding something?

It would have to be a matrix of size `nnodes(mesh), nvariables(equations)` so that `eta` and `v` are contiguous in memory. Alternatively, one could use some types from RecursiveArrayTools.jl, e.g., `ArrayPartition`

JoshuaLampert:
> It would have to be a matrix of size nnodes(mesh), nvariables(equations) so that eta and v are contiguous in memory.

Ah yes. Julia stores column major.

src/equations/bbm_bbm_1d.jl
```diff
@@ -132,8 +132,9 @@ function create_cache(mesh, equations::BBMBBMEquations1D,
                       ::BoundaryConditionPeriodic,
                       RealT, uEltype)
     D = equations.D
-    invImD2 = inv(I - 1 / 6 * D^2 * Matrix(solver.D2))
-    return (invImD2 = invImD2,)
+    invImD2 = cholesky(Symmetric(I - 1 / 6 * D^2 * Matrix(solver.D2)))
+    tmp2 = Array{RealT}(undef, nnodes(mesh))
+    return (invImD2 = invImD2, tmp2 = tmp2)
```
ranocha, line 137: This is still a dense matrix, isn't it? Can you now use the rational operators or at least sparse matrices?

ranocha: IIRC, you were using `sparse` for the results posted at the top of this PR

JoshuaLampert: You mean I should define `invImD2 = cholesky(Symmetric(I - 1 / 6 * D^2 * sparse(solver.D2)))`?

ranocha: Yes

JoshuaLampert: This still gives this error:

>  I tried using `ldiv!(invImD2, deta)`, but this gave the following error message:
> ```julia
> ERROR: LoadError: MethodError: no method matching ldiv!(::SparseArrays.CHOLMOD.Factor{Float64, Int64}, ::SubArr
>```

ranocha: Yeah, this needs the 3-argument version of `ldiv!`

JoshuaLampert: Hm, like in [fe92848](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/108/commits/fe928487ed8a54f02ed91f2bb717a3c9dc66303a)? Still the same error.

ranocha: You would need another temporay array to store the result - it must not alias with the source like in `ldiv!(tmp1, factorization, tmp1)` but has to be independent, e.g., `ldiv!(tmp2, factorization, tmp1)`.

JoshuaLampert: Like [a9a9391](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/108/commits/a9a93914cb5e8859715cc49fca7549ee2ec82d89)? Again the same.

ranocha: Could you please show the full stacktrace of the error?

src/equations/bbm_bbm_1d.jl
```diff
@@ -228,25 +234,27 @@ function rhs!(du_ode, u_ode, t, mesh, equations::BBMBBMEquations1D, initial_cond
...
+        @timeit timer() "dv hyperbolic" tmp2=-solver.D1.plus *
                                               (equations.gravity * eta + 0.5 * v .^ 2)
     else
         @error "unknown type of first-derivative operator: $(typeof(solver.D1))"
     end
 
     @timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations, solver)
 
-    @timeit timer() "deta elliptic" deta[:]=invImD2n * deta
+    @timeit timer() "deta elliptic" ldiv!(invImD2n, tmp1)
```
ranocha, line 250: This is still the 2-argument version of `ldiv!` - did this line lead to the error you mentioned?

JoshuaLampert: No, it shouldn't. This is for the reflecting bc, but up to now I'm only testing the periodic bc.

JoshuaLampert: From the stacktrace above we can see that the error occurs in line 209, which uses the three-argument `ldiv!` version.

src/equations/bbm_bbm_1d.jl
ranocha: Yes, looks like that's it 🤷 

Here is a code snippet that should work (where I have to update the system matrix in each step):
```julia
    system_matrix = Symmetric(...)
    if issparse(system_matrix)
        (; factorization) = parameters
        cholesky!(factorization, system_matrix)
        dv .= factorization \ tmp
    else
        factorization = cholesky!(system_matrix)
        ldiv!(dv, factorization, tmp)
    end
```

JoshuaLampert: Hm, this also doesn't seem to be ideal. I've also just seen that CI is complaining about the system matrix not being positive definite (I can't link to the line because of the GitHub Actions hiccups) and thus the Cholesky decomposition fails. But maybe it is easier (for now?) to go with an LU decomposition instead of a Cholesky decomposition?

JoshuaLampert: In [e68079e](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/108/commits/e68079ee1f4301890e398500bccdd386e072bc8c) I made a proposal for the LU decomposition, which works, has no allocations in the elliptic part and is reasonably fast, I think:
```julia
 ──────────────────────────────────────────────────────────────────────────────
        DispersiveSWE                 Time                    Allocations      
                             ───────────────────────   ────────────────────────
      Tot / % measured:           552ms /  97.8%            627MiB /  99.8%    

 Section             ncalls     time    %tot     avg     alloc    %tot      avg
 ──────────────────────────────────────────────────────────────────────────────
 rhs!                 16.8k    503ms   93.1%  30.0μs    610MiB   97.5%  37.3KiB
   dv hyperbolic      16.8k    132ms   24.4%  7.87μs    338MiB   54.0%  20.6KiB
   deta hyperbolic    16.8k    125ms   23.1%  7.43μs    270MiB   43.2%  16.5KiB
   deta elliptic      16.8k    117ms   21.7%  7.00μs     0.00B    0.0%    0.00B
   dv elliptic        16.8k    113ms   21.0%  6.77μs     0.00B    0.0%    0.00B
   ~rhs!~             16.8k   15.7ms    2.9%   934ns   2.56MiB    0.4%     160B
   source terms       16.8k    586μs    0.1%  34.9ns     0.00B    0.0%    0.00B
 analyze solution       280   37.0ms    6.9%   132μs   15.4MiB    2.5%  56.5KiB
 ──────────────────────────────────────────────────────────────────────────────
```

ranocha: Sounds good 👍

src/equations/bbm_bbm_1d.jl
```diff
@@ -202,9 +208,18 @@ function rhs!(du_ode, u_ode, t, mesh, equations::BBMBBMEquations1D, initial_cond
...
 
-    @timeit timer() "deta elliptic" deta[:]=invImD2 * deta
-    @timeit timer() "dv elliptic" dv[:]=invImD2 * dv
-
+    # To use the in-place version `ldiv!` instead of `\`, we need temporary arrays
+    # since `deta` and `dv` are not stored contiguously
+    @timeit timer() "deta elliptic" begin
+        tmp1[:] = deta
+        ldiv!(tmp3, invImD2, tmp1)
+        deta[:] = tmp3
```
JoshuaLampert, line 216: I set `tmp1` here and not use `tmp1` instead of `deta` already earlier for the hyperbolic part because of the source terms that need to be computed in between and take `dq`, i.e. we already need to set `deta` and `dv` before.

ranocha:
> Another question for my understanding: Do the two error messages `ERROR: LoadError: ArgumentError: in and output vectors must have unit strides` and `ERROR: LoadError: matrix does not have contiguous columns` have the same root cause, namely that we store `deta` and `dv` as a vector instead of a matrix or is the former caused by something else?

Yes, the same root cause

JoshuaLampert:
> Yes, the same root cause

Ok, thanks!

src/equations/bbm_bbm_1d.jl
```diff
@@ -202,9 +208,18 @@ function rhs!(du_ode, u_ode, t, mesh, equations::BBMBBMEquations1D, initial_cond
...
+    end
+    @timeit timer() "dv elliptic" begin
+        tmp2[:] = dv
+        ldiv!(tmp4, invImD2, tmp2)
+        dv[:] = tmp4
```
ranocha, line 221: Then, you would get
```suggestion
        tmp1[:] = dv
        ldiv!(tmp2, invImD2, tmp1)
        dv[:] = tmp2
```

src/equations/bbm_bbm_1d.jl
ranocha: You could get rid of `tmp3, tmp4` and just use `tmp1, tmp2` here, can't you?

src/equations/bbm_bbm_1d.jl
```diff
@@ -241,10 +256,18 @@ function rhs!(du_ode, u_ode, t, mesh, equations::BBMBBMEquations1D, initial_cond
 
     @timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations, solver)
 
-    @timeit timer() "deta elliptic" deta[:]=invImD2n * deta
+    # To use the in-place version `ldiv!` instead of `\`, we need temporary arrays
+    # since `deta` and `dv` are not stored contiguously
+    @timeit timer() "deta elliptic" begin
+        tmp1[:] = deta
+        ldiv!(tmp3, invImD2n, tmp1)
+        deta[:] = tmp3
```
ranocha, line 264: Same here

src/equations/bbm_bbm_1d.jl
```diff
@@ -241,10 +256,18 @@ function rhs!(du_ode, u_ode, t, mesh, equations::BBMBBMEquations1D, initial_cond
...
+    end
     @timeit timer() "dv elliptic" begin
-        dv[2:(end - 1)] = invImD2d * dv[2:(end - 1)]
+        tmp2[:] = dv
+        ldiv!(tmp4, invImD2d, tmp2[2:(end - 1)])
         dv[1] = dv[end] = zero(eltype(dv))
+        dv[2:(end - 1)] = tmp4
```
ranocha, line 270: And here

src/equations/bbm_bbm_1d.jl
JoshuaLampert: Yes, you're right. Thanks! Done in 3da50c61524fbd4d9c5f5a89a2c7bd0e724c515d.

src/equations/bbm_bbm_1d.jl
JoshuaLampert: Here I need another temporary vector because it has a different size (only N - 2 instead of N).

ranocha (approved):

ranocha (approved):

PR #112, opened 2024-05-22
Factorize matrices for `BBMBBMVariableEquations1D`

JoshuaLampert:
Same as #108, but for the variable bathymetry case.

### DIFF
--- a/src/equations/bbm_bbm_variable_bathymetry_1d.jl
+++ b/src/equations/bbm_bbm_variable_bathymetry_1d.jl
@@ -202,14 +202,15 @@ function create_cache(mesh, equations::BBMBBMVariableEquations1D,
     K = Diagonal(D .^ 2)
     if solver.D1 isa PeriodicDerivativeOperator ||
        solver.D1 isa UniformPeriodicCoupledOperator
-        invImDKD = inv(I - 1 / 6 * Matrix(solver.D1) * K * Matrix(solver.D1))
+        invImDKD = lu(I - 1 / 6 * sparse(solver.D1) * K * sparse(solver.D1))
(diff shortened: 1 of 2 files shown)


src/equations/bbm_bbm_variable_bathymetry_1d.jl
```diff
@@ -202,14 +202,15 @@ function create_cache(mesh, equations::BBMBBMVariableEquations1D,
     K = Diagonal(D .^ 2)
     if solver.D1 isa PeriodicDerivativeOperator ||
        solver.D1 isa UniformPeriodicCoupledOperator
-        invImDKD = inv(I - 1 / 6 * Matrix(solver.D1) * K * Matrix(solver.D1))
+        invImDKD = lu(I - 1 / 6 * sparse(solver.D1) * K * sparse(solver.D1))
```
ranocha, line 205:
```suggestion
        sparse_D1 = sparse(solver.D1)
        invImDKD = lu(I - 1 / 6 * sparse_D1 * K * sparse_D1)
```

ranocha (commented):
Just a minor suggestion. It looks like some tolerances and/or reference values may need to be adjusted

src/equations/bbm_bbm_variable_bathymetry_1d.jl
JoshuaLampert: Thanks!

JoshuaLampert:
>  It looks like some tolerances and/or reference values may need to be adjusted

Yes, this was also the case for the other PR. I will do that.

ranocha (approved):
Thanks!

PR #115, opened 2024-05-28
Remove unnecessary part in callback condition

JoshuaLampert:
See https://github.com/trixi-framework/Trixi.jl/pull/1960. This part of the condition is always true and thus unnecessary.


ranocha (approved):
Thanks

PR #120, opened 2024-06-25
Fix failing Cholesky decomposition

JoshuaLampert:
In same cases, the Cholesky decomposition in the Svärd-Kalisch simulation might fail due to non-positive-definiteness coming from too large time steps. This fixes this.

### DIFF
--- a/src/equations/svaerd_kalisch_1d.jl
+++ b/src/equations/svaerd_kalisch_1d.jl
@@ -181,9 +181,15 @@ function create_cache(mesh, equations::SvaerdKalischEquations1D,
        solver.D1 isa UniformPeriodicCoupledOperator
         D1_central = solver.D1
         sparse_D1 = sparse(D1_central)
+        # We use the periodic SBP property
+        #   M * sparse_D1 == -sparse_D1' * M
+        # to avoid possible floating point errors in the symmetry of the matrix.
         minus_MD1betaD1 = sparse_D1' * M * Diagonal(beta_hat) * sparse_D1
     elseif solver.D1 isa PeriodicUpwindOperators
         D1_central = solver.D1.central
+        # We use the periodic upwind SBP property
+        #   M * sparse_D1plus == -sparse_D1minus' * M
+        # to avoid possible floating point errors in the symmetry of the matrix.
         sparse_D1minus = sparse(solver.D1.minus)
         minus_MD1betaD1 = sparse_D1minus' * M * Diagonal(beta_hat) * sparse_D1minus
     else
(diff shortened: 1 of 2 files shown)


src/equations/svaerd_kalisch_1d.jl
```diff
@@ -176,21 +176,22 @@ function create_cache(mesh, equations::SvaerdKalischEquations1D,
     beta_hat = equations.beta * D .^ 3
     gamma_hat = equations.gamma * sqrt.(equations.gravity * D) .* D .^ 3
     tmp2 = similar(h)
+    M = mass_matrix(solver.D1)
     if solver.D1 isa PeriodicDerivativeOperator ||
        solver.D1 isa UniformPeriodicCoupledOperator
         D1_central = solver.D1
         sparse_D1 = sparse(D1_central)
-        D1betaD1 = sparse_D1 * Diagonal(beta_hat) * sparse_D1
+        minus_MD1betaD1 = sparse_D1' * M * Diagonal(beta_hat) * sparse_D1
```
ranocha, line 184:
```suggestion
        # We use the periodic SBP property
        #   M * sparse_D1 == -sparse_D1' * M
        # to avoid possible floating point errors in the symmetry of the matrix.
        minus_MD1betaD1 = sparse_D1' * M * Diagonal(beta_hat) * sparse_D1
```

src/equations/svaerd_kalisch_1d.jl
```diff
@@ -176,21 +176,22 @@ function create_cache(mesh, equations::SvaerdKalischEquations1D,
...
     elseif solver.D1 isa PeriodicUpwindOperators
         D1_central = solver.D1.central
-        D1betaD1 = sparse(solver.D1.plus) * Diagonal(beta_hat) * sparse(solver.D1.minus)
+        sparse_D1minus = sparse(solver.D1.minus)
+        minus_MD1betaD1 = sparse_D1minus' * M * Diagonal(beta_hat) * sparse_D1minus
```
ranocha, line 188:
```suggestion
        # We use the periodic upwind SBP property
        #   M * sparse_D1plus == -sparse_D1minus' * M
        # to avoid possible floating point errors in the symmetry of the matrix.
        sparse_D1minus = sparse(solver.D1.minus)
        minus_MD1betaD1 = sparse_D1minus' * M * Diagonal(beta_hat) * sparse_D1minus
```

ranocha (changes requested):
Thanks! I think a few more comments would be helpful. Other than that, this PR can be merged 👍

JoshuaLampert:
Just to be clear and for later reference: The change that rewrites `hmD1betaD1` be using the periodic (upwind) SBP property is not directly related to the fix of the failing Cholesky decomposition. The fix comes from the check if the Cholesky factorization was successful and setting explicitly rejecting the time step if not. The other change was just in addition to avoid possible floating point errors in the symmetry of the matrix.

ranocha (approved):
Thanks!

PR #127, opened 2024-08-07
implement Serre-Green-Naghdi equations for flat bathymetry

ranocha:
I have implemented the semidiscretization of https://arxiv.org/abs/2408.02665 of the Serre-Green-Naghdi equations for flat bathymetry using central SBP operators. This is the first step and I plan to extend it in the future

- [x] upwind operators
- variable bathymetry (mild-slope approximation and full system)
- hyperbolic approximation

I plan to do this in other PRs to make the individual steps easier to review.


src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+@inline function bathymetry(q, equations::SerreGreenNaghdiEquations1D{BathymetryFlat})
+    return zero(eltype(q))
+end
+
+@inline function waterheight(q, equations::SerreGreenNaghdiEquations1D)
+    return q[1]
+end
+
+# The entropy/energy takes the whole `q` for every point in space
+@inline function energy_total(q, equations::SerreGreenNaghdiEquations1D{BathymetryFlat},
```
JoshuaLampert, line 286: As above, we could think of renaming this `energy_total_modified` and add another quantity `energy_total` computing only the pure shallow water energy part. Even if this is not conserved by the equations, it can be interesting how it evolves under the solution of the Serre-Green-Naghdi equations, I think.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+
+    # `q` is an `ArrayPartition`. It collects the individual arrays for
+    # the water height `h` and the velocity `v`.
+    h, v = q.x
+
+    N = length(v)
+    e = zeros(eltype(q), N)
+
+    # 1/2 g h^2 + 1/2 h v^2 + 1/6 h^3 v_x^2
+    if D isa PeriodicUpwindOperators
```
JoshuaLampert, line 300: See above.

src/semidiscretization.jl
```diff
@@ -159,6 +159,27 @@ function integrate_quantity!(quantity,
     integrate_quantity(func, q, semi)
 end
 
+# The entropy/energy of the Serre-Green-Naghdi equations
+# takes the whole `q` for every point in space since it requires
+# the derivative of the velocity `v_x`.
+function integrate_quantity(func::Union{typeof(energy_total),
```
JoshuaLampert, line 165: Again, if we rename these functions to `*_modified` we would not need the additional methods since they are already covered by the above method.

JoshuaLampert (changes requested):
Thanks a lot for the PR! I left some questions, ideas, and comments below. Some more questions that are in a broader context:
1. What is the plan regarding the full Serre-Green-Naghdi equations and the hyperbolic approximation? Having two separate equations, I guess? To what extend can they share code?
2. One bigger question I already addressed in some places in the code review: What is the plan regarding the different bathymetries? I would like to have only one equation struct covering all types of bathymetry (in contrast to the current state for the BBM-BBM equations). I think in order to have a smooth generalization for other than constant bathymetries in the future, it already pays of to have a clear plan in mind how this generalization should look like. Especially, should we have another variable in the ODE representing the bathymetry? If we do, then this should already be present in the flat bathymetry case, right?
3. From the paper I see there are several possible energy-conserving split forms. You only implemented one, right? I am totally fine with that, but I was only wondering whether you have some plans/ideas to be more flexible?
4. In the paper you also include the possibility to add artificial diffusion? Do you have any plans for integrating this feature in DispersiveShallowWater.jl? If we want to include it, could this be a general (more or less) equations-independent choice and could be switched on/off on the `solver` side rather than the `equation` side? This is definitely not something, we need to decide now, but I already wanted to hear your opinion on that.
5. To have a better impression on the dispersive behaviour of the equations, do you know of an expression for the dispersion relation of the Serre-Green-Naghdi equations (and the hyperbolic approximation). I would be interested in that.
6. I think it would be good to add some more tests. I'm especially thinking of a test, which uses the relaxation callback and tests for exact entropy conservation (I think relaxation does not work out of the box with the current setup because `entropy` needs the whole vector `q` and therefore needs to be handled differently). I think a test for well-balancedness would also be nice (see also my comments below).

Generally, I think a subsequent PR (or more than one) that applies some of your ideas from this PR to the other equations can be useful. I can do that once this PR is merged.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+        inv3 = 1 / 3
+        @. M_h3_3 = inv3 * h^3
+        scale_by_mass_matrix!(M_h3_3, D)
+        system_matrix = Symmetric(Diagonal(M_h)
+                                  +
+                                  Dmat' * Diagonal(M_h3_3) * Dmat)
+    end
+
+    @trixi_timeit timer() "solving elliptic system" begin
+        if issparse(system_matrix)
```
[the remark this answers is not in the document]
JoshuaLampert: Ah, I just realize that for the case of a `FourierDerivativeOperator` the system matrix is not sparse, so this makes sense.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,443 @@
...
+struct BathymetryFlat <: AbstractBathymetry end
+"""
+    bathymetry_flat = DispersiveShallowWater.BathymetryFlat()
+
+A singleton struct indicating a flat bathymetry ``b = 0``.
+"""
+const bathymetry_flat = BathymetryFlat()
+
+@doc raw"""
+    SerreGreenNaghdiEquations1D(gravity)
```
JoshuaLampert, line 11:
```suggestion
    SerreGreenNaghdiEquations1D(; gravity_constant, eta0 = 0.0)
```
The leading semicolon is also missing in the docstrings of the other equations. Could you maybe fix that, too?

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+# TODO: Implement upwind version
+# TODO: Implement variable bathymetry
+function rhs!(dq, q, t, mesh,
+              equations::SerreGreenNaghdiEquations1D{BathymetryFlat},
+              initial_condition,
+              ::BoundaryConditionPeriodic,
+              source_terms::Nothing,
+              solver, cache)
+    if cache.D isa PeriodicUpwindOperators
+        rhs_sgn_flat_upwind!(dq, q, equations, source_terms, cache)
```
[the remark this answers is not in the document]
JoshuaLampert: Upwind operators are already handled in this PR.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,443 @@
...
+            factorization = cholesky!(system_matrix)
+            scale_by_mass_matrix!(tmp, D)
+            ldiv!(dv, factorization, tmp)
+        end
+    end
+
+    return nothing
+end
+
+function rhs_sgn_flat_upwind!(dq, q, equations, source_terms, cache)
```
JoshuaLampert, line 287: Implementing two separate rhs functions for the central and the upwind operators introduces quite some code duplication. Could we maybe make this more elegant? For the other equations I've implemented it in just one function and used branches only for the different cases, where the two versions actually differed. What do you think?

src/equations/serre_green_naghdi_flat_1d.jl
JoshuaLampert: Upwind operators are already implemented in this PR. So this is fine.

examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_upwind.jl
```diff
@@ -0,0 +1,49 @@
...
+
+# create solver with periodic upwind SBP operators of accuracy order 4
+# (note that only central-type with even order are used)
+accuracy_order = 4
+D1 = upwind_operators(periodic_derivative_operator;
+                      derivative_order = 1,
+                      accuracy_order = accuracy_order - 1,
+                      xmin = xmin(mesh), xmax = xmax(mesh),
+                      N = nnodes(mesh))
+D2 = sparse(D1.plus) * sparse(D1.minus) # FIXME: not used
```
JoshuaLampert, line 29: Yeah, good question. We could allow passing `nothing` for `D2` and then pass `nothing` here. But I'm not sure if that's ideal.
I also though about that already when thinking about how to implement KdV for example where we would also need a third derivative operator. I'm open for some nice ideas :smile:

JoshuaLampert: We decided to allow `nothing` and pass it here, but keep it the fact that there is not default for `D2`. Once we need third derivative order operators we also allow to pass `nothing` for it and `D3` will be `nothing` by default.

src/callbacks_step/analysis.jl
```diff
@@ -336,6 +336,16 @@ function analyze(quantity::Union{typeof(energy_total_modified), typeof(entropy_m
     integrate_quantity(quantity, q, semi)
 end
 
+# The entropy/energy of the Serre-Green-Naghdi equations
+# takes the whole `q` for every point in space since it requires
+# the derivative of the velocity `v_x`.
```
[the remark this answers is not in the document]
JoshuaLampert: We decided to call the the functions `energy_total_modified` and `entropy_modified` and keep `entropy` and `energy_total` for the entropy function of the shallow water equations. This should be defined once for all "shallow water type equations" meaning equations, which extend the shallow water equations (i.e. all equations we currently have, but not KdV or BBM for instance). This should be done by another abstract type, which all of the "shallow water type equations" inherit from.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
+abstract type AbstractBathymetry end
+struct BathymetryFlat <: AbstractBathymetry end
```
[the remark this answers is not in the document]
JoshuaLampert: See https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/127#issuecomment-2284393980 2.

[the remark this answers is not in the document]
JoshuaLampert: See https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/127#issuecomment-2284393980 2.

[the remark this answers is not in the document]
JoshuaLampert: It should be possible and would be nice to support.

src/equations/serre_green_naghdi_flat_1d.jl
JoshuaLampert: Since this would mean quite a few of branches and also dead branches, it is easier to keep it as it is.

src/equations/serre_green_naghdi_flat_1d.jl
JoshuaLampert: We decided to rename this to `energy_total_modified`. In this case, this function should get a proper docstring.

src/callbacks_step/analysis.jl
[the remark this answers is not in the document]
JoshuaLampert: This behavior should be documented properly.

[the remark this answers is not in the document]
ranocha: Done in dca91a17927e0f4efedd47c698d23800e24bff9b

src/equations/serre_green_naghdi_flat_1d.jl
ranocha: Done in dca91a17927e0f4efedd47c698d23800e24bff9b

examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_upwind.jl
ranocha: 012303f2917fc0f7ee5aa778976e289924503649

JoshuaLampert (changes requested):

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+       AbstractSerreGreenNaghdiEquations{1, 2}
+    bathymetry::Bathymetry # type of bathymetry
+    gravity::RealT # gravitational constant
+end
+
+function SerreGreenNaghdiEquations1D(; gravity_constant)
+    SerreGreenNaghdiEquations1D(bathymetry_flat, gravity_constant)
+end
+
+varnames(::typeof(prim2prim), ::SerreGreenNaghdiEquations1D) = ("h", "v")
```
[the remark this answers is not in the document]
JoshuaLampert: We decided to put it into the equation as variable.

src/equations/serre_green_naghdi_flat_1d.jl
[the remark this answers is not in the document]
JoshuaLampert: This is still open, @ranocha, right?

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,313 @@
...
+@inline function waterheight_total(q, equations::SerreGreenNaghdiEquations1D)
+    return waterheight(q, equations) + bathymetry(q, equations)
+end
+
+@inline function velocity(q, equations::SerreGreenNaghdiEquations1D)
+    return q[2]
+end
+
+@inline function bathymetry(q, equations::SerreGreenNaghdiEquations1D{BathymetryFlat})
+    return zero(eltype(q))
```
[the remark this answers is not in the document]
JoshuaLampert: Related to the other open convo.

src/equations/serre_green_naghdi_flat_1d.jl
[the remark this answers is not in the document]
JoshuaLampert: Ah no, it is done because the bathymetry is set in the initial condition, which can be constant and non-zero.

src/equations/serre_green_naghdi_flat_1d.jl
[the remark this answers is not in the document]
JoshuaLampert: This is already done.

src/equations/equations.jl
```diff
@@ -210,12 +323,27 @@ Default analysis integrals used by the [`AnalysisCallback`](@ref).
 """
 default_analysis_integrals(::AbstractEquations) = Symbol[]
 
+abstract type AbstractBathymetry end
+struct BathymetryFlat <: AbstractBathymetry end
+"""
+    bathymetry_flat = DispersiveShallowWater.BathymetryFlat()
+
+A singleton struct indicating a flat bathymetry ``b = 0``.
```
JoshuaLampert, line 331:
```suggestion
A singleton struct indicating a flat bathymetry.
```
In our current approach it doesn't need to be 0, does it?

ranocha:
Thanks for your review, @JoshuaLampert! I think we're close to the last iteration 🙂

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,456 @@
...
+# TODO: Implement source terms
+# TODO: Implement variable bathymetry
+function rhs!(dq, q, t, mesh,
+              equations::SerreGreenNaghdiEquations1D{BathymetryFlat},
+              initial_condition,
+              ::BoundaryConditionPeriodic,
+              source_terms::Nothing,
+              solver, cache)
+    if cache.D isa PeriodicUpwindOperators
+        rhs_sgn_flat_upwind!(dq, q, equations, source_terms, cache)
```
JoshuaLampert, line 174:
```suggestion
        rhs_sgn_upwind!(dq, q, equations, source_terms, cache, equations.bathymetry)
```
Would it be easier to dispatch on the bathymetry type rather than including it in the name of the function once we support other bathymetry types, i.e. we would not need to add another `if` clause depending on the bathymetry, right?
(Of course this change would have to lead to other changes, which I don't add here as comments.)

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,456 @@
...
+        rhs_sgn_flat_central!(dq, q, equations, source_terms, cache)
+    end
+
+    return nothing
+end
+
+function rhs_sgn_flat_central!(dq, q, equations, source_terms, cache)
+    # Unpack physical parameters and SBP operator `D` as well as the
+    # SBP operator in sparse matrix form `Dmat`
+    g = equations.gravity
```
JoshuaLampert, line 185:
```suggestion
    g = gravity_constant(equations)
```

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,456 @@
...
+        end
+    end
+
+    return nothing
+end
+
+function rhs_sgn_flat_upwind!(dq, q, equations, source_terms, cache)
+    # Unpack physical parameters and SBP operator `D` as well as the
+    # SBP upwind operator in sparse matrix form `Dmat_minus`
+    g = equations.gravity
```
JoshuaLampert, line 283:
```suggestion
    g = gravity_constant(equations)
```

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,456 @@
...
+        mul!(h_vx_x, D, tmp)
+        # p_+
+        @. tmp = 0.5 * h^2 * (h * v_x + h_x * v) * v_x_upwind
+        mul!(p_x, D_upwind.plus, tmp)
+        # p_0
+        minv6 = -1 / 6
+        @. tmp = minv6 * (h * h2_v_vx_x
+                          +
+                          h^2 * v * h_vx_x)
+        mul!(p_x, D, tmp, 1.0, 1.0)
```
JoshuaLampert, line 321:
```suggestion
        mul!(p_x, D, tmp)
```
Why do you use the 5-argument `mul!` when you pass ones anyway?

test/test_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,68 @@
...
+    @trixi_testset "serre_green_naghdi_soliton.jl" begin
+        @test_trixi_include(joinpath(EXAMPLES_DIR,
+                                     "serre_green_naghdi_soliton.jl"),
+                            tspan=(0.0, 0.1),
+                            l2=[9.994998669268741e-7, 1.4703955789342244e-6, 0.0],
+                            linf=[6.5496216650196e-7, 1.027615853854691e-6, 0.0],
+                            cons_error=[0.0, 8.174581012099225e-10, 0.0],
+                            change_waterheight=0.0,
+                            change_entropy_modified=-3.1093350116861984e-11,
+                            atol=1e-9) # in order to make CI pass
```
JoshuaLampert, line 20: Did you check that we really need the `atol=1e-9` in all the tests?

src/equations/equations.jl
```diff
@@ -133,38 +156,128 @@ See [`momentum`](@ref).
...
 
+"""
+    still_water_surface(q, equations::AbstractShallowWaterEquations)
+
+Return the still water surface ``\\eta_0`` (lake at rest)
+for a given set of `equations`.
+"""
+@inline function still_water_surface(q, equations::AbstractShallowWaterEquations)
+    return equations.eta0
+end
```
JoshuaLampert, line 167: We could consider adding
```julia
varnames(::typeof(still_water_surface), equations) = ("η_0",)
```
and similarly for `bathymetry` further down. If we do this, we should also add them to the `conversion_function`s in the unit tests.

JoshuaLampert (changes requested):
Thanks a lot @ranocha, also for the additional type stability tests. I think this is a big step forward!
I have one final iteration of minor comments/question until this is good to merge.

src/equations/serre_green_naghdi_flat_1d.jl
```diff
@@ -0,0 +1,456 @@
...
+# Discretization that conserves
+# - the total water mass (integral of h) as a linear invariant
+# - the total momentum (integral of h v) as a nonlinear invariant
+# - the total energy
+# for periodic boundary conditions, see
+# - Hendrik Ranocha and Mario Ricchiuto (2024)
+#   Structure-preserving approximations of the Serre-Green-Naghdi
+#   equations in standard and hyperbolic form
+#   [arXiv: 2408.02665](https://arxiv.org/abs/2408.02665)
+# TODO: Implement source terms
```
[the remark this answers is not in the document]
ranocha: Yes, that should work. However, I need to derive a meaningful test first. Thus, I would like to postpone this - since this PR is already quite big. I have added it to the list in #129

src/equations/serre_green_naghdi_flat_1d.jl
ranocha: Since the default values are `1, 0` and not `1, 1`:
```julia
julia> C = [1.0;;]; A = [2.0;;]; B = [3.0;;]; mul!(C, A, B)
1×1 Matrix{Float64}:
 6.0

julia> C = [1.0;;]; A = [2.0;;]; B = [3.0;;]; mul!(C, A, B, 1, 1)
1×1 Matrix{Float64}:
 7.0
```

test/test_serre_green_naghdi_1d.jl
ranocha: No, I didn't - just copy & paste. What's your experience - shall I just remove it and add it back if necessary or shall we try to find an "optimal" value?

src/equations/equations.jl
ranocha: Since we don't use this right now, we may postpone it - or open an issue?

src/equations/serre_green_naghdi_flat_1d.jl
ranocha: 3f0dbfa

src/equations/serre_green_naghdi_flat_1d.jl
JoshuaLampert: Ah, sure. My bad. Sorry.

test/test_serre_green_naghdi_1d.jl
JoshuaLampert: Yes, I would remove it and put it back if necessary. I usually had a look at the differences between the values I got locally (and thus put into the test files) and the values that came from CI and then chose the tolerance to be the smallest value, which is a power of ten for which all tests pass.

ranocha: CI passes with strict tolerances 🥳

JoshuaLampert: That's nice, but why did the test values change in [53a172d](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/127/commits/53a172d90f852befc07b0c4b7484f9935ad4465c)? Are the current values the ones you get locally?

ranocha: Yes, the current values are what I get locally

ranocha: Maybe there was some copy-paste issue earlier or some the implementation changed slightly after writing the tests? 🤷

JoshuaLampert: Alright :+1:

test/test_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,64 @@
...
+        @test_trixi_include(joinpath(EXAMPLES_DIR,
+                                     "serre_green_naghdi_soliton.jl"),
+                            tspan=(0.0, 0.1),
+                            l2=[9.994998669268741e-7, 1.4703973445698635e-6, 0.0],
+                            linf=[6.5496216650196e-7, 1.027617322124641e-6, 0.0],
+                            cons_error=[0.0, 8.174581012099225e-10, 0.0],
+                            change_waterheight=0.0,
+                            change_entropy_modified=-3.1093350116861984e-11)
+
+        @test_allocations(semi, sol, allocs=550_000)
```
JoshuaLampert, line 21: Same question as before: Did you check how many allocation we actually have?

src/equations/svaerd_kalisch_1d.jl
```diff
@@ -1,5 +1,8 @@
 @doc raw"""
-    SvaerdKalischEquations1D(gravity, eta0 = 1.0, alpha = 0.0, beta = 0.2308939393939394, gamma = 0.04034343434343434)
+    SvaerdKalischEquations1D(; gravity_constant, eta0 = 1.0,
```
JoshuaLampert, line 2:
```suggestion
    SvaerdKalischEquations1D(; gravity_constant, eta0 = 0.0,
```

test/test_serre_green_naghdi_1d.jl
ranocha: I made them a bit stricter if possible, rounding them to the next reasonable multiple of `50_000`. For example, the first test has `523_304` allocations locally (rounded to `550_000`) and the second test has `393_360` allocations locally, rounded up to `450_000` (since it was too close to `400_000`).

JoshuaLampert: Sounds good.

JoshuaLampert (approved):
Thanks again.

Issue #129, opened 2024-08-09
Roadmap for Serre-Green-Naghdi


ranocha:
- [ ] flat bathymetry with central operators: #127
- [ ] flat bathymetry with upwind operators
- [ ] variable bathymetry (mild-slope approximation and full system)
- [ ] hyperbolic approximation
- [ ] #126

Issue #130, opened 2024-08-12
Include different types of bathymetries


JoshuaLampert:
In #127 we decided to define different types of bathymetries (flat, variable, mild slope) as structs, which are used to dispatch. In any case, we keep the bathymetry as variable in the equation to be able to possibly extend the package in the future to also deal with time-dependent bathymetries. This concept should also be used for the existing equations:

- [x] BBM-BBM: Currently we have two different equation structs. They should be combined into one, which also holds the bathymetry type. They only support the flat and the variable case (not the mild slope case), see #147
- [x] Svärd-Kalisch: Add another field representing the bathymetry type. For a flat bathymetry, we should also implement a more special semidiscretization, which utilizes the constant bathymetry to allow for higher-order derivative operator since in this case the coefficients are also constant and go in front of the higher-order derivative. (#146 for the introduction of the `bathymetry_type`, flat bathymetry is still missing)

ranocha:
> > Another question, which just came into my mind: What do we do if you pass bathymetry_flat to the equation, but then set a non-constant bathymetry in the initial condition. I haven't thought about that before. Should this raise a warning or an error or nothing?

> Where could we add such a check? My first idea was to look at create_cache, but that doesn't have the initial condition

It could live in `Semidiscretization`, but then I would only add such a check when all equations have a field `bathymetry` (or `bathymetry_type`), i.e. after finishing #130.

_Originally posted by @JoshuaLampert in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/issues/135#issuecomment-2293798516_

JoshuaLampert:
When all `AbstractShallowWaterEquations1D` have a `bathymetry_type` and have $\eta$, $v$ and $b$ as variables in the ODE, we can also make more functions dispatch on `AbstractShallowWaterEquations1D` rather than on the different equations, e.g., `varnames`,  `initial_condition_dingemans`, and all the conversion functions like `prim2cons`, `cons2prim` etc.

JoshuaLampert:
We could think of making `bathymetry_variable` the default for all equations.

ranocha:
Closed by #147

Issue #122, opened 2024-07-27
Use sparse Cholesky decomposition with `ldiv!`


JoshuaLampert:
Currently, we use an LU factorization for the `BBMBBMEquations1D` and `BBMBBMVariableEquations1D` although the matrix is spd, which means we could use a Cholesky decomposition. We did this because `ldiv!` was not supported for sparse Cholesky factorizations, see the discussion in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/108#discussion_r1608314166. For the `SvaerdKalischEquations1D` we use a sparse Cholesky decomposition, but use `\` instead of `ldiv!`, again, because `ldiv!` was not implemented for sparse Cholesky decompositions.
Once https://github.com/JuliaSparse/SparseArrays.jl/pull/547 (thanks for the PR @ranocha!) is available, it should be possible to use a sparse Cholesky decomposition together with `ldiv!` for all equations.

ranocha:
The PR is merged and shall be included in Julia v1.12 (see https://github.com/JuliaSparse/SparseArrays.jl/pull/547#issuecomment-2272994661)

JoshuaLampert:
In this case, we could reuse the function `solve_system_matrix!`, which is currently only used for the Serre-Green-Naghdi equations for the others, too.

PR #135, opened 2024-08-15
variable bathymetry Serre-Green-Naghdi equations

ranocha:
This is the next step of #129


src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
+@doc raw"""
```
JoshuaLampert, line 1: It's a bit hard to review this file when it is renamed. I would like to see a proper diff. Can you maybe rename this file back to "serre_green_naghdi_flat_1d.jl" for the moment until I reviewed it and in the end rename it back? Or alternatively, we could rename the file in a separate PR.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+function energy_total_modified(q_global,
+                               equations::SerreGreenNaghdiEquations1D,
+                               cache)
+    # unpack physical parameters and SBP operator `D`
+    g = equations.gravity
+    (; D, v_x) = cache
+
+    # `q_global` is an `ArrayPartition`. It collects the individual arrays for
+    # the water height `h` and the velocity `v`.
+    h, v = q_global.x
```
JoshuaLampert, line 928:
```suggestion
    eta, v = q_global.x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        (; b, b_x) = cache
+        @. b = equations.eta0 - q_global.x[3]
+        if D isa PeriodicUpwindOperators
+            mul!(b_x, D.central, b)
+        else
+            mul!(b_x, D, b)
+        end
+    end
+
+    @. e = 1 / 2 * g * h^2 + 1 / 2 * h * v^2 + 1 / 6 * h * (-h * v_x + 1.5 * v * b_x)^2
```
JoshuaLampert, line 954:
```suggestion
    @. e = 1 / 2 * g * eta^2 + 1 / 2 * h * v^2 + 1 / 6 * h * (-h * v_x + 1.5 * v * b_x)^2
```
Otherwise this is wrong, isn't it? You would also need to compute h before. Can you please check that `eta` and `h` are used consistently everywhere?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+Return the modified total energy of the primitive variables `q_global` for the
+[`SerreGreenNaghdiEquations1D`](@ref).
+It contains an additional term containing a
+derivative compared to the usual [`energy_total`](@ref) modeling
+non-hydrostatic contributions. The [`energy_total_modified`](@ref)
+is a conserved quantity (for periodic boundary conditions).
+
+For a [`bathymetry_flat`](@ref) the total energy is given by
+```math
+\\frac{1}{2} g h^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
```
JoshuaLampert, line 913:
```suggestion
\\frac{1}{2} g \\eta^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+    # Unpack physical parameters and SBP operator `D` as well as the
+    # SBP upwind operator in sparse matrix form `Dmat_minus`
+    g = gravity_constant(equations)
+    (; Dmat_minus) = cache
+    D_upwind = cache.D
+    D = D_upwind.central
+
+    # `q` and `dq` are `ArrayPartition`s. They collect the individual
+    # arrays for the water height `h` and the velocity `v`.
+    h, v = q.x
```
JoshuaLampert, line 490:
```suggestion
    eta, v = q.x
```
This would be `\eta` usually. Of course the remaining part of this function needs to be adjusted, too.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+
+function rhs_sgn_central!(dq, q, equations, source_terms, cache, ::BathymetryFlat)
+    # Unpack physical parameters and SBP operator `D` as well as the
+    # SBP operator in sparse matrix form `Dmat`
+    g = gravity_constant(equations)
+    (; D, Dmat) = cache
+
+    # `q` and `dq` are `ArrayPartition`s. They collect the individual
+    # arrays for the water height `h` and the velocity `v`.
+    h, v = q.x
```
JoshuaLampert, line 390:
```suggestion
    eta, v = q.x
```

JoshuaLampert (changes requested):
Only a preliminary review with a question regarding the use of `eta` and `h`.

JoshuaLampert:
Another question, which just came into my mind: What do we do if you pass `bathymetry_flat` to the `equation`, but then set a non-constant bathymetry in the initial condition. I haven't thought about that before. Should this raise a warning or an error or nothing?

test/test_serre_green_naghdi_1d.jl
JoshuaLampert, line 1: Could you please define `bathymetry` in all elixirs and then simply pass `bathymetry = bathymetry_mild_slope` (or `bathymetry_variable`) to `@test_trixi_include` instead of passing the whole `equation`?

JoshuaLampert (commented):
Can you please also add a well-balancedness test?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+    if D isa PeriodicUpwindOperators
+        v_x_upwind = zero(h)
+        p_0 = zero(h)
+
+        Dmat_minus = sparse(D.minus)
+
+        # Floating point errors accumulate a bit and the system matrix
+        # is not necessarily perfectly symmetric but only up to
+        # round-off errors. We wrap it here to avoid issues with the
+        # factorization.
```
JoshuaLampert, line 274: This comment looks like it is a bit out of place, isn't it?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+                     h2_v_vx_x, h_vx_x, p_h, p_x, tmp,
+                     M_h_p_h_bx2, M_h3_3, M_h2_bx,
+                     D, Dmat)
+        else
+            Dmat = sparse(D)
+
+            # Floating point errors accumulate a bit and the system matrix
+            # is not necessarily perfectly symmetric but only up to
+            # round-off errors. We wrap it here to avoid issues with the
+            # factorization.
```
JoshuaLampert, line 317: See above

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+            cache = (; h, h_x, v_x, h_hpb_x, b, b_x, hv_x, v2_x,
+                     h2_v_vx_x, h_vx_x, p_h, p_x, tmp,
+                     M_h_p_h_bx2, M_h3_3, M_h2_bx,
+                     D, Dmat, factorization)
+        end
+    end
+
+    if equations.bathymetry isa BathymetryVariable
+        ψ = zero(h)
+        cache = (; cache..., ψ)
```
JoshuaLampert, line 351:
```suggestion
        psi = zero(h)
        cache = (; cache..., psi)
```
I usually follow the style guide of Trixi.jl to not include any non-ASCII characters.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+    dh, dv, dD = dq.x
+    fill!(dD, zero(eltype(dD)))
+
+    @trixi_timeit timer() "hyperbolic terms" begin
+        # Compute all derivatives required below
+        (; h, h_x, v_x, h_hpb_x, b, b_x, hv_x, v2_x,
+        h2_v_vx_x, h_vx_x, p_h, p_x, tmp,
+        M_h_p_h_bx2, M_h3_3, M_h2_bx) = cache
+        if equations.bathymetry isa BathymetryVariable
+            (; ψ) = cache
```
JoshuaLampert, line 598:
```suggestion
            (; psi) = cache
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        @. p_h = (0.5 * h * (h * v_x + h_x * v) * v_x
+                  -
+                  inv6 * h2_v_vx_x
+                  -
+                  inv6 * h * v * h_vx_x)
+        @. tmp = h * b_x * v^2
+        mul!(p_x, D, tmp)
+        @. p_h += 0.25 * p_x
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ = 0.125 * p_x
```
JoshuaLampert, line 628:
```suggestion
            @. psi = 0.125 * p_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        end
+        @. tmp = b_x * v
+        mul!(p_x, D, tmp)
+        @. p_h += 0.25 * h * v * p_x
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ += 0.125 * h * v * p_x
```
JoshuaLampert, line 634:
```suggestion
            @. psi += 0.125 * h * v * p_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        end
+        @. p_h = p_h - 0.25 * (h_x * v + h * v_x) * b_x * v
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ = ψ - 0.125 * (h_x * v + h * v_x) * b_x * v
```
JoshuaLampert, line 638:
```suggestion
            @. psi = psi - 0.125 * (h_x * v + h * v_x) * b_x * v
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+                   -
+                   0.5 * v^2 * h_x
+                   +
+                   0.5 * hv_x * v
+                   -
+                   0.5 * h * v * v_x
+                   + p_x
+                   + 1.5 * p_h * b_x)
+        if equations.bathymetry isa BathymetryVariable
+            @. tmp = tmp - ψ * b_x
```
JoshuaLampert, line 664:
```suggestion
            @. tmp = tmp - psi * b_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        end
+
+        @. b = equations.eta0 - q.x[3]
+        @. h = eta - b
+        mul!(b_x, D, b)
+
+        mul!(h_x, D, h)
+        mul!(v_x, D, v)
+        mul!(v_x_upwind, D_upwind.minus, v)
+        @. tmp = h * (h + b)
```
JoshuaLampert, line 749:
```suggestion
        @. tmp = h * eta
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+    dh, dv, db = dq.x
+    fill!(db, zero(eltype(db)))
+
+    @trixi_timeit timer() "hyperbolic terms" begin
+        # Compute all derivatives required below
+        (; h, h_x, v_x, v_x_upwind, h_hpb_x, b, b_x, hv_x, v2_x,
+        h2_v_vx_x, h_vx_x, p_h, p_0, p_x, tmp,
+        M_h_p_h_bx2, M_h3_3, M_h2_bx) = cache
+        if equations.bathymetry isa BathymetryVariable
+            (; ψ) = cache
```
JoshuaLampert, line 739:
```suggestion
            (; psi) = cache
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        # p_0
+        minv6 = -1 / 6
+        @. p_h = minv6 * (h2_v_vx_x
+                          +
+                          h * v * h_vx_x)
+        @. tmp = h * b_x * v^2
+        mul!(p_x, D, tmp)
+        @. p_h += 0.25 * p_x
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ = 0.125 * p_x
```
JoshuaLampert, line 769:
```suggestion
            @. psi = 0.125 * p_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        end
+        @. tmp = b_x * v
+        mul!(p_x, D, tmp)
+        @. p_h += 0.25 * h * v * p_x
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ += 0.125 * h * v * p_x
```
JoshuaLampert, line 775:
```suggestion
            @. psi += 0.125 * h * v * p_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+        end
+        @. p_0 = p_h * h
+        mul!(p_x, D, p_0)
+        # p_+
+        @. tmp = (0.5 * h * (h * v_x + h_x * v) * v_x_upwind
+                  -
+                  0.25 * (h_x * v + h * v_x) * b_x * v)
+        if equations.bathymetry isa BathymetryVariable
+            @. ψ = 0.125 * (h_x * v + h * v_x) * b_x * v
```
JoshuaLampert, line 784:
```suggestion
            @. psi = 0.125 * (h_x * v + h * v_x) * b_x * v
```
Another question: Do I miss something or do you set $\psi$ already a few lines above, but never use it and here you set it again without using the $\psi$ from before?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+                   -
+                   0.5 * v^2 * h_x
+                   +
+                   0.5 * hv_x * v
+                   -
+                   0.5 * h * v * v_x
+                   + p_x
+                   + 1.5 * p_h * b_x)
+        if equations.bathymetry isa BathymetryVariable
+            @. tmp = tmp - ψ * b_x
```
JoshuaLampert, line 811:
```suggestion
            @. tmp = tmp - psi * b_x
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+end
+
+@inline function bathymetry(q, equations::SerreGreenNaghdiEquations1D)
+    D = q[3]
+    return equations.eta0 - D
+end
+
+@inline function waterheight(q, equations::SerreGreenNaghdiEquations1D)
+    return waterheight_total(q, equations) - bathymetry(q, equations)
+end
```
JoshuaLampert, line 898: This function does not depend on the concrete equation, right? So could we put it in `src/equations/equations.jl` and dispatch on `AbstractShallowWaterEquations` (and remove it from the other files, too)?

src/equations/serre_green_naghdi_1d.jl
ranocha: Done in #136

ranocha:
> Another question, which just came into my mind: What do we do if you pass `bathymetry_flat` to the equation, but then set a non-constant bathymetry in the initial condition. I haven't thought about that before. Should this raise a warning or an error or nothing?

Where could we add such a check? My first idea was to look at `create_cache`,  but that doesn't have the initial condition

src/equations/serre_green_naghdi_1d.jl
ranocha: Yes, we could - but it assumes that we use the variables `eta, v, D` since we would have more efficient ways to compute it otherwise. However, since you wanted to use these variables for consistency (which makes this PR so annoying to review), I will move it 👍

test/test_serre_green_naghdi_1d.jl
ranocha: I see that it would make the test file look simpler - but it would make the examples look more complicated. Since the examples are the user-facing part and the tests are mainly for developers, it may be nice to keep the examples simple?

JoshuaLampert: On the other hand I can also image a user who would like to run some elixir with another bathymetry type easily by using `trixi_include` and thus including something like `bathymetry = bathymetry_variable` would make their life easier.

JoshuaLampert:
> > Another question, which just came into my mind: What do we do if you pass bathymetry_flat to the equation, but then set a non-constant bathymetry in the initial condition. I haven't thought about that before. Should this raise a warning or an error or nothing?

> Where could we add such a check? My first idea was to look at create_cache, but that doesn't have the initial condition

It could live in `Semidiscretization`, but then I would only add such a check when all equations have a field `bathymetry` (or `bathymetry_type`), i.e. after finishing #130.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,980 @@
...
+  Contribution â l'étude des écoulements permanents et variables dans les canaux
+  [DOI: 10.1051/lhb/1953034](https://doi.org/10.1051/lhb/1953034)
+- Green and Naghdi (1976)
+  A derivation of equations for wave propagation in water of variable depth
+  [DOI: 10.1017/S0022112076002425](https://doi.org/10.1017/S0022112076002425)
+
+The semidiscretization implemented here conserves
+- the total water mass (integral of h) as a linear invariant
+- the total momentum (integral of h v) as a nonlinear invariant
+- the total energy
```
ranocha, line 63:
```suggestion
- the total modified energy
```

JoshuaLampert:
Now we have a bunch of conflicts to solve... :roll_eyes:

ranocha:
> Now we have a bunch of conflicts to solve... 🙄

Done

JoshuaLampert:
Thanks!

test/test_serre_green_naghdi_1d.jl
```diff
@@ -59,6 +149,96 @@ EXAMPLES_DIR = joinpath(examples_dir(), "serre_green_naghdi_1d")
...
+        @test_trixi_include(joinpath(EXAMPLES_DIR,
+                                     "serre_green_naghdi_well_balanced.jl"),
+                            tspan=(0.0, 2.0),
+                            l2=[0, 0, 0],
+                            linf=[0, 0, 0],
+                            cons_error=[0, 0, 0],
+                            change_waterheight=0.0,
+                            change_momentum=0.0,
+                            change_entropy_modified=0.0,
+                            lake_at_rest=0.0)
```
JoshuaLampert, line 193: Out of curiosity: In reality we get values around machine precision and not exactly 0, right?

JoshuaLampert (changes requested):
Thanks a lot @ranocha! This looks already quite good. Please find some minor comments below.
I also have two more questions and one request:
1. Maybe a strange one: But why didn't the test values change after [various fixes](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/135/commits/2ae501231a2af4c1988aed59f049902608701b5b)? Since there were some true changes in the numerics (the psi fix and the fix in the `entropy_modified`) I would have assumed this should have changed the test values. Is it because the changes are so small that they are within the tolerance? And if so, can you maybe run the elixirs again and update the values with the new values?
2. Did you perform some performance tests similar to section 11.10 in your preprint and compared it to the code from your paper? I would be happy if we could reach a similar performance here than you report in the paper.
3. Can you report a convergence test with the soliton solution?

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
```diff
@@ -0,0 +1,49 @@
...
+boundary_conditions = boundary_condition_periodic
+
+# create homogeneous mesh
+coordinates_min = -140.0
+coordinates_max = 100.0
+N = 512
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic upwind SBP operators of accuracy order 4
+# (note that only central-type with even order are used)
```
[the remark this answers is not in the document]
ranocha: Yes, exactly 👍

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -28,26 +59,36 @@ References for the Serre-Green-Naghdi system can be found in
...
     eta0::RealT # constant still-water surface
 end
 
-function SerreGreenNaghdiEquations1D(; gravity_constant, eta0 = 0.0)
-    eta0 == 0.0 ||
-        @warn "The still-water surface needs to be 0 for the Serre-Green-Naghdi equations"
-    SerreGreenNaghdiEquations1D(bathymetry_flat, gravity_constant, eta0)
+function SerreGreenNaghdiEquations1D(bathymetry = bathymetry_flat;
+                                     gravity_constant, eta0 = 0.0)
+    SerreGreenNaghdiEquations1D(bathymetry, gravity_constant, eta0)
```
ranocha, line 81:
```suggestion
function SerreGreenNaghdiEquations1D(bathymetry_type = bathymetry_flat;
                                     gravity_constant, eta0 = 0.0)
    SerreGreenNaghdiEquations1D(bathymetry_type, gravity_constant, eta0)
```

test/test_serre_green_naghdi_1d.jl
ranocha: That's what I would expect in general but here we get

```julia
julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_well_balanced.jl");
[...]

────────────────────────────────────────────────────────────────────────────────────────────────────
Simulation running 'SerreGreenNaghdiEquations1D-BathymetryVariable' with 'initial_condition_discontinuous_well_balancedness'
────────────────────────────────────────────────────────────────────────────────────────────────────
 #timesteps:                 60                run time:       3.35273791e-01 s
 Δt:             5.00000000e-01                └── GC time:    1.25329125e-01 s (37.381%)
 sim. time:      3.00000000e+01 (100.000%)
 #DOF:                      512                alloc'd memory:        249.217 MiB

 Variable:       η                v                D             
 L2 error:       0.00000000e+00   0.00000000e+00   0.00000000e+00
 Linf error:     0.00000000e+00   0.00000000e+00   0.00000000e+00
 |∫q - ∫q₀|:     0.00000000e+00   0.00000000e+00   0.00000000e+00

 Integrals:    
 ∫η          :   4.00000000e+00
 ∫P          :   0.00000000e+00
 ∫U_modified :   4.00000000e+00
 ∫|η-η₀|     :   0.00000000e+00
[...]
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,960 @@
...
+    end
+
+    @trixi_timeit timer() "assembling elliptic operator" begin
+        if equations.bathymetry isa BathymetryMildSlope
+            factor = 0.75
+        else # equations.bathymetry isa BathymetryVariable
+            factor = 1.0
+        end
+        # The code below is equivalent to
+        #   dv .= (Diagonal(h .+ factor .* h .* b_x.^2) - Dmat * (Diagonal(1/3 .* h.^3) * Dmat - Diagonal(0.5 .* h.^2 .* b_x)) - Diagonal(0.5 .* h.^2 .* b_x) * Dmat) \ tmp
```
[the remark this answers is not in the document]
ranocha: It is fixed in the current version

src/equations/serre_green_naghdi_1d.jl
ranocha: c1cbda4

src/equations/equations.jl
```diff
@@ -325,14 +342,39 @@ Default analysis integrals used by the [`AnalysisCallback`](@ref).
...
+See also [`bathymetry_mild_slope`](@ref) and [`bathymetry_variable`](@ref).
 """
 const bathymetry_flat = BathymetryFlat()
 
+struct BathymetryMildSlope <: AbstractBathymetry end
+"""
+    bathymetry_mild_slope = DispersiveShallowWater.BathymetryMildSlope()
+
+A singleton struct indicating a variable bathymetry with mild-slope
+approximation.
```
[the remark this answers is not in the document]
ranocha: Done in 33115e3

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -187,14 +389,18 @@ function rhs_sgn_central!(dq, q, equations, source_terms, cache, ::BathymetryFla
...
+    dh, dv, dD = dq.x # dh = deta since b is constant in time
+    fill!(dD, zero(eltype(dD)))
 
     @trixi_timeit timer() "hyperbolic terms" begin
         # Compute all derivatives required below
-        (; h_x, v_x, h2_x, hv_x, v2_x, h2_v_vx_x,
+        (; h, b, h_x, v_x, h2_x, hv_x, v2_x, h2_v_vx_x,
         h_vx_x, p_x, tmp, M_h, M_h3_3) = cache
 
+        @. b = equations.eta0 - q.x[3]
```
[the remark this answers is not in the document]
ranocha: I introduce `D1` as SBP op. consistently in 5578e539e745bde2eb0272c89423f14cd74753fd

ranocha:
> 3\. Can you report a convergence test with the soliton solution?

```julia
julia> convergence_test("examples/serre_green_naghdi_1d/serre_green_naghdi_soliton.jl", 5, N = 64)

l2
64   1.25e-01  -         64   3.54e-01  -         64   0.00e+00  -         
128  1.43e-02  3.12      128  3.97e-02  3.15      128  0.00e+00  NaN       
256  9.90e-04  3.85      256  2.74e-03  3.86      256  0.00e+00  NaN       
512  6.23e-05  3.99      512  1.73e-04  3.99      512  0.00e+00  NaN       
1024 3.29e-06  4.24      1024 9.13e-06  4.24      1024 0.00e+00  NaN       

mean           3.80      mean           3.81      mean           NaN       
linf
64   5.74e-02  -         64   1.61e-01  -         64   0.00e+00  -         
128  7.23e-03  2.99      128  1.94e-02  3.05      128  0.00e+00  NaN       
256  5.07e-04  3.83      256  1.36e-03  3.83      256  0.00e+00  NaN       
512  3.19e-05  3.99      512  8.64e-05  3.98      512  0.00e+00  NaN       
1024 1.63e-06  4.29      1024 4.44e-06  4.28      1024 0.00e+00  NaN       

mean           3.78      mean           3.79      mean           NaN   

julia> convergence_test("examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_upwind.jl", 5, N = 64)

l2
64   1.25e-01  -         64   3.55e-01  -         64   0.00e+00  -         
128  1.43e-02  3.13      128  3.99e-02  3.15      128  0.00e+00  NaN       
256  9.78e-04  3.87      256  2.74e-03  3.86      256  0.00e+00  NaN       
512  6.21e-05  3.98      512  1.75e-04  3.97      512  0.00e+00  NaN       
1024 3.85e-06  4.01      1024 1.08e-05  4.02      1024 0.00e+00  NaN       

mean           3.75      mean           3.75      mean           NaN       
linf
64   5.72e-02  -         64   1.61e-01  -         64   0.00e+00  -         
128  7.26e-03  2.98      128  1.95e-02  3.04      128  0.00e+00  NaN       
256  5.18e-04  3.81      256  1.43e-03  3.78      256  0.00e+00  NaN       
512  3.38e-05  3.94      512  9.37e-05  3.93      512  0.00e+00  NaN       
1024 1.94e-06  4.12      1024 5.49e-06  4.09      1024 0.00e+00  NaN       

mean           3.71      mean           3.71      mean           NaN 
```

Both elixirs use an accuracy order of 4.

ranocha:
> 1. Maybe a strange one: But why didn't the test values change after [various fixes](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/135/commits/2ae501231a2af4c1988aed59f049902608701b5b)? Since there were some true changes in the numerics (the psi fix and the fix in the `entropy_modified`) I would have assumed this should have changed the test values. Is it because the changes are so small that they are within the tolerance? And if so, can you maybe run the elixirs again and update the values with the new values?

I re-ran the DIngemans setup with the current code and updated the test values - only the conservation error of the momentum changed slightly (within tolerances).

test/test_serre_green_naghdi_1d.jl
JoshuaLampert: Ok, I remember that I had something similar once when I forgot to set $\eta_0\neq 0$, which means everything is just exactly zero all the time, but didn't really test true well-balancedness. This is not the case here, but do you have another idea why this is exactly zero?

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
```diff
@@ -0,0 +1,49 @@
...
+bathymetry_type = bathymetry_variable # or bathymetry_mild_slope
+equations = SerreGreenNaghdiEquations1D(bathymetry_type;
+                                        gravity_constant = 9.81)
+
+initial_condition = initial_condition_dingemans
+boundary_conditions = boundary_condition_periodic
+
+# create homogeneous mesh
+coordinates_min = -140.0
+coordinates_max = 100.0
```
JoshuaLampert, line 17: These are different to the Dingemans elixirs of the other equations. I see you used these values in your preprint. Do you think it makes sense take the same interval bounds across all Dingemans elixirs for better comparability (either use $[-138, 46]$ everywhere or $[-140, 100]$ everywhere)?

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
```diff
@@ -0,0 +1,49 @@
+using OrdinaryDiffEq
+using DispersiveShallowWater
+using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
+
+###############################################################################
+# Semidiscretization of the Svärd-Kalisch equations
```
JoshuaLampert, line 6:
```suggestion
# Semidiscretization of the Serre-Green-Naghdi equations
```

examples/serre_green_naghdi_1d/serre_green_naghdi_well_balanced.jl
```diff
@@ -0,0 +1,66 @@
+using OrdinaryDiffEq
+using DispersiveShallowWater
+using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
+
+###############################################################################
+# Semidiscretization of the BBM-BBM equations
```
JoshuaLampert, line 6:
```suggestion
# Semidiscretization of the Serre-Green-Naghdi equations
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -28,26 +59,36 @@ References for the Serre-Green-Naghdi system can be found in
...
+    bathymetry_type::Bathymetry # type of bathymetry
     gravity::RealT # gravitational constant
     eta0::RealT # constant still-water surface
 end
 
-function SerreGreenNaghdiEquations1D(; gravity_constant, eta0 = 0.0)
-    eta0 == 0.0 ||
-        @warn "The still-water surface needs to be 0 for the Serre-Green-Naghdi equations"
-    SerreGreenNaghdiEquations1D(bathymetry_flat, gravity_constant, eta0)
+function SerreGreenNaghdiEquations1D(bathymetry_type = bathymetry_flat;
```
JoshuaLampert, line 79: Honest question: Do you think it's best to make `bathymetry_flat` the default rather than `bathymetry_variable`? I know that it would be breaking to change this, but would it make sense to have the most general case the default?

ranocha:
> 2\. Did you perform some performance tests similar to section 11.10 in your preprint and compared it to the code from your paper? I would be happy if we could reach a similar performance here than you report in the paper.

That lead me down a rabbit hole 😅 Along the way, I found a quite nasty bug (5a4dbb2265b568b459e21df11cd7128e61045391) that manifested for second-order operators. With the fix, I get the following results

```julia
julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl", io = devnull, bathymetry_type = bathymetry_variable, N = 1_000);
[ Info: You just called `trixi_include`. Julia may now compile the code, please be patient.

julia> @benchmark solve(ode, Tsit5(); save_everystep = false, callback = callbacks)
BenchmarkTools.Trial: 26 samples with 1 evaluation.
 Range (min … max):  190.737 ms … 196.548 ms  ┊ GC (min … max): 3.77% … 4.10%
 Time  (median):     193.543 ms               ┊ GC (median):    3.69%
 Time  (mean ± σ):   193.423 ms ±   1.745 ms  ┊ GC (mean ± σ):  3.67% ± 0.23%

  █ ▁ ▁       ▁▁▁▁ █    ▁    ▁ █ ▁█ ▁▁ ▁   ▁ ▁             ▁▁ █  
  █▁█▁█▁▁▁▁▁▁▁████▁█▁▁▁▁█▁▁▁▁█▁█▁██▁██▁█▁▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁██▁█ ▁
  191 ms           Histogram: frequency by time          197 ms <

 Memory estimate: 684.67 MiB, allocs estimate: 67864.

julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl", io = devnull, bathymetry_type = bathymetry_variable, N = 2_000);
[ Info: You just called `trixi_include`. Julia may now compile the code, please be patient.

julia> @benchmark solve(ode, Tsit5(); save_everystep = false, callback = callbacks)
BenchmarkTools.Trial: 13 samples with 1 evaluation.
 Range (min … max):  381.811 ms … 436.045 ms  ┊ GC (min … max): 2.50% … 4.59%
 Time  (median):     409.998 ms               ┊ GC (median):    3.44%
 Time  (mean ± σ):   406.086 ms ±  17.010 ms  ┊ GC (mean ± σ):  3.40% ± 0.64%

  █▁         ▁               ▁  ▁█ ▁  ▁▁             ▁        ▁  
  ██▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁██▁█▁▁██▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁█ ▁
  382 ms           Histogram: frequency by time          436 ms <

 Memory estimate: 1.36 GiB, allocs estimate: 72824.

julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl", io = devnull, bathymetry_type = bathymetry_variable, N = 3_000);
[ Info: You just called `trixi_include`. Julia may now compile the code, please be patient.

julia> @benchmark solve(ode, Tsit5(); save_everystep = false, callback = callbacks)
BenchmarkTools.Trial: 9 samples with 1 evaluation.
 Range (min … max):  540.357 ms … 621.959 ms  ┊ GC (min … max): 1.34% … 1.74%
 Time  (median):     597.424 ms               ┊ GC (median):    1.82%
 Time  (mean ± σ):   580.736 ms ±  33.005 ms  ┊ GC (mean ± σ):  1.85% ± 0.28%

  █                                                              
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▇▁▁▇▁▁▁▁▇▁▁▁▁▁▁▁▇ ▁
  540 ms           Histogram: frequency by time          622 ms <

 Memory estimate: 2.04 GiB, allocs estimate: 84813.

julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl", io = devnull, bathymetry_type = bathymetry_variable, N = 4_000);
[ Info: You just called `trixi_include`. Julia may now compile the code, please be patient.

julia> @benchmark solve(ode, Tsit5(); save_everystep = false, callback = callbacks)
BenchmarkTools.Trial: 7 samples with 1 evaluation.
 Range (min … max):  816.021 ms … 858.928 ms  ┊ GC (min … max): 1.67% … 1.60%
 Time  (median):     825.590 ms               ┊ GC (median):    1.66%
 Time  (mean ± σ):   829.784 ms ±  14.942 ms  ┊ GC (mean ± σ):  1.68% ± 0.08%

  ██       █   █    █             █                           █  
  ██▁▁▁▁▁▁▁█▁▁▁█▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  816 ms           Histogram: frequency by time          859 ms <

 Memory estimate: 2.71 GiB, allocs estimate: 84801.

julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl", io = devnull, bathymetry_type = bathymetry_variable, N = 5_000);
[ Info: You just called `trixi_include`. Julia may now compile the code, please be patient.

julia> @benchmark solve(ode, Tsit5(); save_everystep = false, callback = callbacks)
BenchmarkTools.Trial: 5 samples with 1 evaluation.
 Range (min … max):  1.006 s …   1.042 s  ┊ GC (min … max): 1.70% … 1.98%
 Time  (median):     1.019 s              ┊ GC (median):    1.73%
 Time  (mean ± σ):   1.022 s ± 13.656 ms  ┊ GC (mean ± σ):  1.82% ± 0.24%

  █                █  █              █                    █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.01 s         Histogram: frequency by time        1.04 s <

 Memory estimate: 3.39 GiB, allocs estimate: 84821.
```

Thus, the results are roughly comparable (a little bit slower but often within standard deviation).

JoshuaLampert (changes requested):
Some more smaller suggestions and comments.
In general we have some code duplication. Would it be possible and wise to put some recurring patterns into functions, e.g., thing like
```julia
        if equations.bathymetry_type isa BathymetryMildSlope
            factor = 0.75
        else # equations.bathymetry_type isa BathymetryVariable
            factor = 1.0
        end
```

JoshuaLampert:
> Thus, the results are roughly comparable (a little bit slower but often within standard deviation).

The results look good. Thanks a lot!

examples/serre_green_naghdi_1d/serre_green_naghdi_conservation.jl
```diff
@@ -0,0 +1,69 @@
...
+#   Structure-preserving approximations of the Serre-Green-Naghdi
+#   equations in standard and hyperbolic form
+#   [arXiv: 2408.02665](https://arxiv.org/abs/2408.02665)
+
+using OrdinaryDiffEq
+using DispersiveShallowWater
+using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
+
+###############################################################################
+# Semidiscretization of the Svärd-Kalisch equations
```
JoshuaLampert, line 14:
```suggestion
# Semidiscretization of the Serre-Green-Naghdi equations
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -97,59 +181,191 @@ function create_cache(mesh,
...
+    M_h2_bx = zero(h)
+
+    # b_x appears in the system matrix, so it must not be zero to get the
+    # correct pattern of the factorization. Just setting it to unity does
+    # also not work since some terms cancel. Thus, b_x needs to be
+    # "variable enough" to get the correct pattern of non-zero entries in
+    # the system matrix and the factorization.
+    # TODO: This is a hack and should be improved. It would be ideal if we
+    #       had access to the initial condition and the exact value of the
+    #       bathymetry here.
```
JoshuaLampert, line 275: Wait...don't we _have_ access to the initial condition, cf. line 245?

JoshuaLampert, line 275: Was this only a bug that influenced the new conservation elixir or does it also influence any of the other examples (maybe the well-balancedness test does not return plain zeros anymore)?

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
ranocha: I would leave that up to you - which version would you like to have?

test/test_serre_green_naghdi_1d.jl
ranocha: No further idea, sorry

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
JoshuaLampert: To be able say which version I prefer it would be interesting to know why you used other values in the first place? To avoid wave interaction between the waves entering the periodic domain in the left and some small waves, which still existed in the left area from the initial wave train?

src/equations/serre_green_naghdi_1d.jl
ranocha: The bug only influenced second-order methods as far as I have observed - and we used higher-order methods. The well-balanced test case still returns exactly 0

test/test_serre_green_naghdi_1d.jl
JoshuaLampert: Ok, I guess it's just fine then. Having _too_ good results isn't too bad :sweat_smile:

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: Ok, thanks for the information.

ranocha: But if we evaluated it here, we would evaluate it twice. Moreover, for the case of time-dependent bathymetry that you have in mind, the initial bathymetry may still be constant so that we need a hack like this.

src/equations/serre_green_naghdi_1d.jl
ranocha: Let's see whether we need other breaking changes when introducing the hyperbolic approximation? I'm fine making the most general (but most expensive) case the default

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
ranocha: Yes, exactly

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: Ok, we can keep it like this for now, but maybe change this in the future when we have a breaking change anyway. We will have a breaking change anyway once we add the `bathymetry_type` also to the other equations and thus merge `BBMBBMEquations1D` and `BBMBBMVariableEquations1D` in one equation. I'm planning to do this once we are finished with integrating SGN.

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
JoshuaLampert: Did you observe any influence of these interactions? At least visually it looks like the waves from the initial wave train are really small also in the smaller domain. And also in the larger domain you have waves entering the left again. I think if the influence of the interactions are not too big I would keep the smaller domain to have a better resolution.

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: I see. In that case I also don't know of any good solution.

ranocha:
> Some more smaller suggestions and comments. In general we have some code duplication. Would it be possible and wise to put some recurring patterns into functions, e.g., thing like
> 
> ```julia
>         if equations.bathymetry_type isa BathymetryMildSlope
>             factor = 0.75
>         else # equations.bathymetry_type isa BathymetryVariable
>             factor = 1.0
>         end
> ```

Done in c7ff29a129e56ec43da45bb35944eb98caa9473c and 0062720c18c8e60a7e1970f333fc5ec52f373497

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -97,62 +181,204 @@ function create_cache(mesh,
...
+    @show initial_condition
```
JoshuaLampert, line 261:
```suggestion

```

examples/serre_green_naghdi_1d/serre_green_naghdi_dingemans.jl
ranocha: Done in a0cdae4

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -97,62 +181,204 @@ function create_cache(mesh,
...
+                     D1mat' * (Diagonal(M_h3_3) * D1mat
+                               -
+                               Diagonal(M_h2_bx))
+                     -
+                     Diagonal(M_h2_bx) * D1mat)
+end
+
+function solve_system_matrix!(dv, system_matrix, rhs,
+                              ::SerreGreenNaghdiEquations1D,
+                              D1, cache)
```
JoshuaLampert, line 362: This looks quite nice. For the future we could even think of reusing this function also for the other equations. For the Svärd-Kalisch equations I think we could also use it already, but not for BBM-BBM equations because we use LU instead of Cholesky decomposition, which will hopefully change some day, see #122.

JoshuaLampert, line 362: This also changes the `cache` (via the `factorization`), right? I think the convention is to put the variables, which are mutated first, right? This would mean that `cache` has to come before the equations and `D1`.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -342,36 +553,240 @@ function rhs_sgn_upwind!(dq, q, equations, source_terms, cache, ::BathymetryFlat
...
+                   0.5 * h * v * v_x
+                   + p_x
+                   + 1.5 * p_h * b_x)
+        if equations.bathymetry_type isa BathymetryVariable
+            @. tmp = tmp - psi * b_x
+        end
+    end
+
+    # The code below is equivalent to
+    #   dv .= (Diagonal(h .+ 0.75 .* h .* b_x.^2) - D1mat * (Diagonal(1/3 .* h.^3) * D1mat - Diagonal(0.5 .* h.^2 .* b_x)) - Diagonal(0.5 .* h.^2 .* b_x) * D1mat) \ tmp
```
JoshuaLampert, line 777:
```suggestion
    #   dv .= (Diagonal(h .+ factor .* h .* b_x.^2) - D1mat * (Diagonal(1/3 .* h.^3) * D1mat - Diagonal(0.5 .* h.^2 .* b_x)) - Diagonal(0.5 .* h.^2 .* b_x) * D1mat) \ tmp
```

test/test_serre_green_naghdi_1d.jl
```diff
@@ -59,6 +149,121 @@ EXAMPLES_DIR = joinpath(examples_dir(), "serre_green_naghdi_1d")
...
+    end
+
+    @trixi_testset "serre_green_naghdi_conservation.jl" begin
+        @test_trixi_include(joinpath(EXAMPLES_DIR,
+                                     "serre_green_naghdi_conservation.jl"),
+                            l2=[1.3655498085989206, 2.3967486930606716, 0.0],
+                            linf=[1.001076318001934, 0.8052527556023067, 0.0],
+                            cons_error=[0.0, 0.0002674927404067162, 0.0],
+                            change_entropy=-0.0584189861183404,
+                            change_entropy_modified=0.059273537492344985)
```
JoshuaLampert, line 250: Why do we have so big changes in the entropy? Is this expected?

JoshuaLampert (changes requested):
Final iteration. This look already very nice!

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: I have tracked this in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/issues/130#issuecomment-2294812733. So I'm resolving this conversation.

src/equations/serre_green_naghdi_1d.jl
ranocha: Yes, but `rhs!` also changes the `cache`. In Trixi.jl, we adopted the convention that it's okay to put the `cache` last if it's not the primary thing that is changed.

test/test_serre_green_naghdi_1d.jl
ranocha: It's running for a bit longer time to stress the conservation test (compilation time is dominating anyway when running the tests). Moreover, we use less strict tolerances here (to make them comparable to the benchmarks we used in the paper). If I set `abstol = 1.0e-7, reltol = 1.0e-7`, I get the lower errors close to the other elixirs, e.g., `-3.6345863918541e-7` for `tspan = (0.0, 1.0)`.

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: Makes sense. I thought about this because in `assemble_system_matrix!` you put `cache` as first argument, but there it _is_ the primary thing that is changed. So that's fine. Thanks for the clarification!

test/test_serre_green_naghdi_1d.jl
JoshuaLampert: Ok, then it's fine :+1:

JoshuaLampert: Did you make a quick check that you get the same values as in Table 3 (c) -- (f) from your paper?

ranocha: A random sample of two values say yes

JoshuaLampert: Perfect, thanks!

JoshuaLampert (approved):
LGTM! Thanks a lot!

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: We just keep it as it is for this PR?

JoshuaLampert:
Now we only have this nasty codecov error on MacOS. But since it seems to fail consistently, I would like to find a fix before merging this PR.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -423,7 +833,15 @@ is a conserved quantity (for periodic boundary conditions).
...
 For a [`bathymetry_flat`](@ref) the total energy is given by
 ```math
-\\frac{1}{2} g h^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
+\\frac{1}{2} g \\eta^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
+```
+For a [`bathymetry_mild_slope`](@ref) the total energy is given by
+```math
+\\frac{1}{2} g \\eta^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h (-h v_x + 1.5 v b_x)^2.
+```
+For a [`bathymetry_variable`](@ref) the total energy has the additional term
```
JoshuaLampert, line 842:
```suggestion
For a [`bathymetry_variable`](@ref) the total modified energy has the additional term
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -423,7 +833,15 @@ is a conserved quantity (for periodic boundary conditions).
 
 For a [`bathymetry_flat`](@ref) the total energy is given by
 ```math
-\\frac{1}{2} g h^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
+\\frac{1}{2} g \\eta^2 + \\frac{1}{2} h v^2 + \\frac{1}{6} h^3 v_x^2.
+```
+For a [`bathymetry_mild_slope`](@ref) the total energy is given by
```
JoshuaLampert, line 838:
```suggestion
For a [`bathymetry_mild_slope`](@ref) the total modified energy is given by
```
The same is also a few lines above (where I cannot comment).

JoshuaLampert (changes requested):

JoshuaLampert (approved):

JoshuaLampert (approved):
Thanks!

PR #139, opened 2024-08-17
hyperbolic Serre-Green-Naghdi equations

ranocha:
Next step of #129

### TODO

- [x] Discuss name clashes and decide how to proceed
- [x] Format
- [x] Add tests and more elixirs
- [x] How to initialize variables? Ideally, I would like to use `w = -(D1 * v)` instead of having to specify a value manually in the initial condition.
- [x] Dingemans
- [x] Improve performance for flat bathymetry (and test both for the soliton setup)


src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,328 @@
...
+    HyperbolicSerreGreenNaghdiEquations1D(bathymetry_type, gravity_constant, eta0)
+end
+
+# TODO: How shall we handle eta=h+b vs the new variable η?
+function varnames(::typeof(prim2prim), ::HyperbolicSerreGreenNaghdiEquations1D)
+  return ("η", "v", "D", "w", "η")
+end
+function varnames(::typeof(prim2cons), ::HyperbolicSerreGreenNaghdiEquations1D)
+  return ("h", "hv", "b", "hw", "hη")
+end
```
ranocha, line 81: @JoshuaLampert How would you like to handle the name clash of the total water height $$\eta = h + b$$ and the new variable coming from the hyperbolic relaxation approach $$\eta \approx h$$?

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,364 @@
...
+#       However, we cannot use that name since it's not an analytical solution.
+#       How shall we handle this?
+"""
+    initial_condition_soliton(x, t, equations::HyperbolicSerreGreenNaghdiEquations1D, mesh)
+
+A soliton solution of the [`SerreGreenNaghdiEquations1D`](@ref)
+used for convergence tests in a periodic domain.
+
+See also [`initial_condition_convergence_test`](@ref).
+"""
```
ranocha, line 95: Another name clash

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,364 @@
...
+It contains additional terms compared to the usual [`energy_total`](@ref)
+modeling non-hydrostatic contributions. The `energy_total_modified`
+is a conserved quantity (for periodic boundary conditions).
+
+For a [`bathymetry_mild_slope`](@ref) (and a [`bathymetry_flat`](@ref)),
+the total modified energy is given by
+```math
+\\frac{1}{2} g \\eta^2 + \\frac{1}{2} h v^2 +
+\\frac{1}{6} h w^2 + \\frac{\\lambda}{6} h (1 - \\eta / h)^2.
+```
```
ranocha, line 339: In this case, it's not strictly required to get the full vector

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -828,7 +827,7 @@ Return the modified total energy of the primitive variables `q_global` for the
 [`SerreGreenNaghdiEquations1D`](@ref).
 It contains an additional term containing a
 derivative compared to the usual [`energy_total`](@ref) modeling
-non-hydrostatic contributions. The [`energy_total_modified`](@ref)
+non-hydrostatic contributions. The `energy_total_modified`
```
ranocha, line 830: This is the docstring for `energy_total_modified`, so it doesn't really make sense to link to it, I think

JoshuaLampert: I agree. Thanks!

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -381,7 +380,7 @@ end
 # Discretization that conserves
 # - the total water mass (integral of h) as a linear invariant
 # - the total momentum (integral of h v) as a nonlinear invariant
-# - the total energy
+# - the total modified energy
```
JoshuaLampert, line 383: Can you add one line above that the total momentum is only conserved for constant bathymetry?

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,364 @@
...
+
+`q_global` is a vector of the primitive variables at ALL nodes.
+`cache` needs to hold the SBP operators used by the `solver`.
```
JoshuaLampert, line 342:
```suggestion
```
This is not needed in this case, right? See also above.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Ok, I think that's fine. See also my suggestion below.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Good question. I would be fine keeping it like it is now (maybe with a bit more explanations in the docstring that this function is the same as `initial_condition_convergence_test` for `SerreGreenNaghdiEquations1D`, but is not an analytical solution of `HyperbolicSerreGreenNaghdiEquations1D`). I think it's not too bad that these two initial conditions are not called the same.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: I already feared this question reading your paper :smile:
I don't know what the best answer is, but maybe call it $H$ (I think we don't have $H$ right now) and clarify in docstrings what this is exactly?

JoshuaLampert (commented):
Thanks! A preliminary review regarding your comments/questions.
I saw some non-ASCII code characters in executable code. Could you replace them by ASCII code characters, please?
>  How to initialize variables? Ideally, I would like to use w = -(D1 * v) instead of having to specify a value manually in the initial condition.

I also thought about that. I agree that this would be really nice. I think it would be ideal if both would be possible, i.e., defining an initial condition consisting of three variables with defaults for $w$ and $\eta$ ($H$ or whatever) or defining an initial condition consisting of all five values. I haven't thought this through, but would it somehow be possible to fiddle around with `compute_coefficients` by adding some dispatching on `HyperbolicSerreGreenNaghdiEquations1D` and some branching whether the output of `initial_condition` is three or five elements long?

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,364 @@
...
+The gravitational constant is denoted by `g` and the bottom topography
+(bathymetry) ``b = \eta_0 - D``. The water height above the bathymetry
+is therefore given by ``h = \eta - \eta_0 + D``.
+The total water height is therefore given by ``\eta = h + b``.
+
+Two types of variable `bathymetry_type` are supported:
+- [`bathymetry_flat`](@ref): flat bathymetry (typically ``b = 0`` everywhere)
+- [`bathymetry_mild_slope`](@ref): variable bathymetry with mild-slope approximation
+
+For the mild-slope approximation, the Serre-Green-Naghdi equations are
```
JoshuaLampert, line 28: This system corresponds to a mild-slope approximation of the bathymetry? From the preprint I would have expected that this is for a general variable bathymetry and the mild-slope approximation was only applied to the full system. If this system indeed describes a mild-slope approximation, does there also exist a formulation for a general bathymetry without the mild-slope assumption?

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,364 @@
...
+  describing long free surface gravity waves
+  [DOI: 10.1088/1361-6544/aa712d](https://doi.org/10.1088/1361-6544/aa712d)
+- Busto, Dumbser, Escalante, Favrie, and Gavrilyuk.
+  On High Order ADER Discontinuous Galerkin Schemes for First Order Hyperbolic
+  Reformulations of Nonlinear Dispersive Systems
+  [DOI: 10.1007/s10915-021-01429-8](https://doi.org/10.1007/s10915-021-01429-8)
+
+The semidiscretization implemented here conserves
+- the total water mass (integral of h) as a linear invariant
+- the total modified energy
```
JoshuaLampert, line 52:
```suggestion
- the total water mass (integral of h) as a linear invariant
- the total momentum (integral of h v) as a nonlinear invariant if the bathymetry is constant
- the total modified energy
```
Correct me if I'm wrong.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
ranocha: Well, they represent the same physics, so it's a bit annoying but I don't know a better solution right now 🤷

src/equations/hyperbolic_serre_green_naghdi_1d.jl
ranocha: It does not conserve the total momentum - the new non-hydrostatic pressure terms involving $$\eta$$ ($$H$$?) are problematic

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -381,7 +380,7 @@ end
 # Discretization that conserves
 # - the total water mass (integral of h) as a linear invariant
 # - the total momentum (integral of h v) as a nonlinear invariant
```
ranocha, line 382:
```suggestion
# - the total momentum (integral of h v) as a nonlinear invariant for flat bathymetry
```

src/equations/hyperbolic_serre_green_naghdi_1d.jl
ranocha: The papers of Gavrilyuk et al. that we looked at only developed an approximation for flat bathymetry (Favrie and Gavrilyuk) with mild-slope approximation. That's why we got a mild-slope version for the original SGN equations when translating the split form from the hyperbolic approximation.

ranocha:
> I saw some non-ASCII code characters in executable code. Could you replace them by ASCII code characters, please?

I used them on purpose to make it easier to search and replace stuff after deciding how to call the new variables

ranocha:
```julia
julia> convergence_test("examples/hyperbolic_serre_green_naghdi_1d/hyperbolic_serre_green_naghdi_manufactured.jl", 4);
[...]
l2
η                        v                        D                        w                        H                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   5.65e-02  -         50   3.46e-02  -         50   0.00e+00  -         50   1.17e+00  -         50   5.85e-02  -         
100  3.61e-03  3.97      100  2.25e-03  3.94      100  0.00e+00  NaN       100  7.51e-02  3.96      100  3.73e-03  3.97      
200  2.26e-04  4.00      200  1.42e-04  3.99      200  0.00e+00  NaN       200  4.71e-03  3.99      200  2.34e-04  4.00      
400  1.42e-05  4.00      400  1.02e-05  3.79      400  0.00e+00  NaN       400  2.95e-04  4.00      400  1.46e-05  4.00      

mean           3.99      mean           3.91      mean           NaN       mean           3.99      mean           3.99      
linf
η                        v                        D                        w                        H                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   1.21e-01  -         50   5.21e-02  -         50   0.00e+00  -         50   2.19e+00  -         50   1.27e-01  -         
100  7.81e-03  3.95      100  3.36e-03  3.95      100  0.00e+00  NaN       100  1.41e-01  3.96      100  8.59e-03  3.88      
200  4.91e-04  3.99      200  2.16e-04  3.96      200  0.00e+00  NaN       200  8.83e-03  3.99      200  5.38e-04  4.00      
400  3.15e-05  3.96      400  2.33e-05  3.21      400  0.00e+00  NaN       400  5.51e-04  4.00      400  3.37e-05  4.00      

mean           3.97      mean           3.71      mean           NaN       mean           3.98      mean           3.96      

```

src/equations/hyperbolic_serre_green_naghdi_1d.jl
ranocha: I have chosen `H`

ranocha:
Including some optimizations for flat bathymetry:

```julia
julia> q = copy(ode.u0); dq = similar(q); @benchmark $(ode.f)($dq, $q, $(ode.p), $0.0)
BenchmarkTools.Trial: 10000 samples with 4 evaluations.
 Range (min … max):  7.864 μs …  13.969 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.927 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   8.165 μs ± 647.504 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▇▆▂▂▃▃           ▂▁▂▁▂                     ▁               ▂
  █████████▅▅▄▅▁▄▃█████████▇▆▅▅▆▃▆▆▆▇██▇▇▇▇▆▇▇███▇▅▅▄▆▆▅▅▃▅▆▆ █
  7.86 μs      Histogram: log(frequency) by time        11 μs <

 Memory estimate: 304 bytes, allocs estimate: 5.

julia> redirect_stdio(stdout = devnull, stderr = devnull) do
           trixi_include("examples/hyperbolic_serre_green_naghdi_1d/hyperbolic_serre_green_naghdi_soliton.jl", interval = 1000, bathymetry_type = bathymetry_mild_slope);
       end;

julia> q = copy(ode.u0); dq = similar(q); @benchmark $(ode.f)($dq, $q, $(ode.p), $0.0)
BenchmarkTools.Trial: 10000 samples with 3 evaluations.
 Range (min … max):  8.597 μs …  17.208 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.917 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   9.159 μs ± 774.910 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇▇▂ █▇▇▇▂               ▁▁▁▁▁▁ ▁            ▁               ▂
  ███▆██████▇███▇▇█▇█▇██████████████▇█▇▇▇▆██▇███▇▇▇▆▆▆▅▆▅▆▅▅▅ █
  8.6 μs       Histogram: log(frequency) by time      12.2 μs <

 Memory estimate: 16 bytes, allocs estimate: 2.
```

test/test_hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,190 @@
...
+                                0.0,
+                                0.006470346460325516,
+                                0.0003406400402869991,
+                            ],
+                            cons_error=[0.0,
+                                1.204145701494186e-5,
+                                0.0,
+                                1.6023987782225921,
+                                4.456003127373265e-6],
+                            atol=1e-8) # to make CI pass
```
ranocha, line 155:
```suggestion
                            atol=1e-7) # to make CI pass
```

JoshuaLampert (changes requested):
This looks already really nice! I think we are not far from merging this. I especially like your trait system to allow all or only a reduced set of variables in the initial condition. I knew you would come up with a very nice idea for implementing that what was very fuzzy in my head 🙂

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Ok, thanks for the clarification!

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Yes, you're right. I think I looked at the wrong theorem yesterday.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Thanks!

src/equations/hyperbolic_serre_green_naghdi_1d.jl
JoshuaLampert: Yeah, it' not perfect, but I can live with in (at least for now). The name `initial_condition_convergence_test` doesn't _describe_ the physics, but rather the fact of it being an analytical solution of some equations, so it's not _that_ confusing, I think.

JoshuaLampert:
The convergence test and benchmark look nice. Thanks!

test/test_hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,190 @@
+module TestSerreGreenNaghdi1D
```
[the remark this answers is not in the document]
JoshuaLampert: Looks like I made the same copy/paste error in for the visualization and Svärd-Kalisch tests. Can you adjust the names of the `module`s there as well, please?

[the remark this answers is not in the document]
ranocha: Done in [2688626](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/139/commits/2688626b1977528b5daa480787ed8e4d2f00f028)

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,476 @@
...
+        end
+    end
+
+    @trixi_timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations,
+                                                       solver)
+
+    return nothing
+end
+
+@inline function prim2cons(q, equations::HyperbolicSerreGreenNaghdiEquations1D)
```
JoshuaLampert, line 414: Would it maybe make sense to add another conversion function, which just returns the first three variables of the primitive variables, something like `prim2physical` (or similar)? I imagine sometimes one might be interested only in these and then one could use the conversion, e.g. when plotting by passing `conversion_function = prim2physical` to `plot`. For the other equations this function would then simply be 
```julia
prim2physical(q, equations::AbstractShallowWaterEquations) = prim2prim(q, equations)
```

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,476 @@
...
+        else
+            @. dv = -(g * h_hpb_x - g * (h + b) * h_x
+                      +
+                      0.5 * h * v2_x - 0.5 * v^2 * h_x
+                      +
+                      0.5 * hv_x * v - 0.5 * h * v * v_x
+                      + lambda_6 * (H_over_h * H_over_h * h_x - H2_h_x)
+                      + lambda_3 * (1 - H_over_h) * H_x
+                      + lambda_2 * (1 - H_over_h) * b_x) / h
+        end
```
JoshuaLampert, line 384:
```suggestion
         @. dv = -(g * h_hpb_x - g * (h + b) * h_x
                   +
                   0.5 * h * v2_x - 0.5 * v^2 * h_x
                   +
                   0.5 * hv_x * v - 0.5 * h * v * v_x
                   + lambda_6 * (H_over_h * H_over_h * h_x - H2_h_x)
                   + lambda_3 * (1 - H_over_h) * H_x) / h
        if !(bathymetry_type isa BathymetryFlat)
            @. dv += (lambda_2 * (1 - H_over_h) * b_x) / h
        end
```
Could we make this shorter to reduce code duplication?

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,476 @@
...
+                   -
+                   0.5 * h * w * v_x) + lambda * (1 - H_over_h)) / h
+
+        # No special split form for energy conservation required:
+        # H_t + v H_x + 3/2 v b_x = w
+        if bathymetry_type isa BathymetryFlat
+            @. dH = -v * H_x + w
+        else
+            @. dH = -v * H_x - 1.5 * v * b_x + w
+        end
```
JoshuaLampert, line 405:
```suggestion
        @. dH = -v * H_x + w
        if !(bathymetry_type isa BathymetryFlat)
            @. dH -= 1.5 * v * b_x
        end
```
Similar to above.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,476 @@
...
+end
+
+@inline function cons2prim(u, equations::HyperbolicSerreGreenNaghdiEquations1D)
+    h, hv, b = u
+
+    eta = h + b
+    v = hv / h
+    D = equations.eta0 - b
+    w = u[4] / h
+    H = u[5] / h
```
JoshuaLampert, line 432:
```suggestion
    h, hv, b, hw, hH = u

    eta = h + b
    v = hv / h
    D = equations.eta0 - b
    w = hw / h
    H = hH / h
```

JoshuaLampert (commented):
Looks mostly good! I have only some minor comments/suggestions/questions.

examples/hyperbolic_serre_green_naghdi_1d/hyperbolic_serre_green_naghdi_manufactured.jl
```diff
@@ -0,0 +1,43 @@
...
+tspan = (0.0, 1.0)
+ode = semidiscretize(semi, tspan)
+summary_callback = SummaryCallback()
+analysis_callback = AnalysisCallback(semi; interval = 1000,
+                                     extra_analysis_errors = (:conservation_error,),
+                                     extra_analysis_integrals = (waterheight_total,
+                                                                 momentum, entropy))
+callbacks = CallbackSet(analysis_callback, summary_callback)
+
+sol = solve(ode, Tsit5(), abstol = 1e-9, reltol = 1e-9,
```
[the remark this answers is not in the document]
ranocha: If you go to very strict tolerances, a higher-order method can become more efficient. This may not necessarily be the case here but for example with tolerances `1.0e-12`

src/equations/hyperbolic_serre_green_naghdi_1d.jl
ranocha: Should we make `prim2physical` the default for plotting?

JoshuaLampert: Yes sounds good.

docs/src/overview.md
```diff
@@ -146,7 +146,7 @@ always plot to first subplot) and `plot_initial`. You can also provide a `conver
 of the primitive variables `q` at one node, and the `equations` as input and should return an `SVector` of any length as output. For a user defined conversion function,
 there should also exist a function `varnames(conversion, equations)` that returns a `Tuple` of the variable names used for labelling. The conversion function can, e.g.,
 be [`prim2cons`](@ref) or [`waterheight_total`](@ref) if one only wants to plot the total water height. The resulting plot will have one subplot for each of the returned
-variables of the conversion variable. By default, the conversion function is just [`prim2prim`](@ref), i.e. the identity.
+variables of the conversion variable. By default, the conversion function is just [`prim2phys`](@ref), i.e., the identity for most equations.
```
JoshuaLampert, line 149:
```suggestion
variables of the conversion variable. By default, the conversion function is just [`prim2phys`](@ref), which computes the physical variables
from the primitive ones, i.e., the identity for most equations.
```
For a bit more clarity.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -0,0 +1,481 @@
...
+    @. H = eta + D - equations.eta0 # h = (h + b) + (eta0 - b) - eta0
+
+    # w ≈ -h v_x
+    mul!(w, D1, v)
+    @. w = -H * w
+
+    return nothing
+end
+
+# TODO: There is another name clash. For the SerreGreenNaghdiEquations1D,
```
JoshuaLampert, line 149:
```suggestion
# TODO: There is name clash. For the SerreGreenNaghdiEquations1D,
```

JoshuaLampert (changes requested):
Thanks a lot! Once my final two suggestions are resolved, this is ready to merge.

JoshuaLampert (approved):
Thanks!

PR #142, opened 2024-08-18
Run CI on Julia v1.10

JoshuaLampert:
Since Julia v1.11 will be out probably soon and v1.10 will be the new LTS, I would like to run CI also on v1.10.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -44,7 +44,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.9'
+          - '1.10'
         os:
           - ubuntu-latest
           - macos-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,6 +51,10 @@ jobs:
           - windows-latest
         arch:
           - x64
+        include:
+          - version: '1.9'
+            os: ubuntu-latest
+            arch: x64
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2


ranocha:
Sounds reasonable. However, we still support Julia v1.9 - and I would like to keep it this way unless you have a very good reason not to do so. My motivation for this is mainly that there are people out there who are not power users of Julia like we, so they will likely have older versions of it. I would like to support them. That's something I have experienced with BSeries.jl.

Thus, we could for example keep one job on Julia v1.9 and move the others to Julia v1.10.

JoshuaLampert:
> Thus, we could for example keep one job on Julia v1.9 and move the others to Julia v1.10.

Sounds good. I added a test, which runs with v1.9 on Ubuntu.

JoshuaLampert:
What about the Downgrade job? I would run it on v1.10. Do you agree?

ranocha:
> What about the Downgrade job? I would run it on v1.10. Do you agree?

I don't have a strong opinion there. One could argue that the oldest versions of all dependencies will likely be available on the oldest Julia versions, though. But as I said, I'm fine either way

JoshuaLampert:
Ok, let's keep it on v1.9 for now.

JoshuaLampert:
Sorry, I mixed up the PRs. I wanted to merge the other one...

JoshuaLampert:
Are you fine with this PR @ranocha? If yes, I would create a new release. If not, I would first consider your suggestion before creating a new release.
Sorry again for not waiting for your review :grimacing:

ranocha (commented):
Fine with me 👍

Issue #11, opened 2023-06-07
Allocations in `RelaxationCallback`


ranocha:
https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/55b6cc61825b89cd3661497520a391589763eed4/src/callbacks_step/relaxation.jl#L82-L84

This will allocate new vectors whenever it's called. That's of course totally fine for now but may slow down the process a bit when you use bigger state vectors (increased grid refinement).

PR #146, opened 2024-08-21
Adapt Svärd-Kalisch equations to Serre-Green-Naghdi equations

JoshuaLampert:
In this PR I try to make the implementation of the Svärd-Kalisch equations follow a more similar style as in the Serre-Green-Naghdi equations. This includes:

- having a `bathymetry_type` as additional field. Up to now, only a variable bathymetry is implemented (as we had before). It might be reasonable to also add the possibility to have a `bathymetry_flat` and then use the higher-order derivative operators for the dispersive terms (this, however, would require to have a third-derivative operator in the `solver`). I would like to keep this for a separate PR in the future.
- avoid any allocations in the hyperbolic part. I pre-allocated the necessary data structures in `create_cache` and overwrite them in `rhs!` now instead of always allocating them again. Now we have 0 allocations for the hyperbolic part.
- unify comments and docstrings a bit.
- reuse `solve_system_matrix!` from the Serre-Green-Naghdi equations.
- make some smaller fixes and simplification along the way.

Additionally, I avoided the allocations in `energy_total_modified` by reusing `tmp1`. I don't know if this is a good idea on the other hand. This seems dangerous to me since `tmp1` will be changed. This also changes the returned value, right? Should I revert this change and allocate a new vector each time again, @ranocha?

### DIFF
--- a/src/callbacks_step/analysis.jl
+++ b/src/callbacks_step/analysis.jl
@@ -298,7 +298,7 @@ function analyze_integrals!(io, current_integrals, i, analysis_integrals::NTuple
     quantity = first(analysis_integrals)
     remaining_quantities = Base.tail(analysis_integrals)
 
-    res = analyze(quantity, q, t, semi)
+    res = analyze!(semi, quantity, q, t)
     current_integrals[i] = res
     @printf(io, " %-12s:", pretty_form_utf(quantity))
     @printf(io, "  % 10.8e", res)
@@ -326,16 +326,7 @@ function (cb::DiscreteCallback{Condition, Affect!})(sol) where {Condition,
     return (; l2 = l2_error, linf = linf_error)
 end
 
-function analyze(quantity, q, t, semi::Semidiscretization)
-    integrate_quantity(q -> quantity(q, semi.equations), q, semi)
-end
-
-# The modified entropy/energy of the Svärd-Kalisch and
-# Serre-Green-Naghdi equations
-# takes the whole `q` for every point in space since it requires
-# the derivative of the velocity `v_x`.
-function analyze(quantity::Union{typeof(energy_total_modified), typeof(entropy_modified)},
-                 q, t, semi::Semidiscretization)
+function analyze!(semi::Semidiscretization, quantity, q, t)
     integrate_quantity!(semi.cache.tmp1, quantity, q, semi)
 end
 
--- a/src/callbacks_step/analysis.jl
+++ b/src/callbacks_step/analysis.jl
@@ -336,7 +336,7 @@ end
 # the derivative of the velocity `v_x`.
 function analyze(quantity::Union{typeof(energy_total_modified), typeof(entropy_modified)},
                  q, t, semi::Semidiscretization)
-    integrate_quantity(quantity, q, semi)
+    integrate_quantity!(semi.cache.tmp1, quantity, q, semi)
 end
 
 pretty_form_utf(::typeof(waterheight_total)) = "∫η"
--- a/src/equations/equations.jl
+++ b/src/equations/equations.jl
@@ -270,12 +270,20 @@ contributions.
 `q_global` is a vector of the primitive variables at ALL nodes.
 `cache` needs to hold the SBP operators used by the `solver` if non-hydrostatic
 terms are present.
+
+Internally, this function allocates a vector for the output and
+calls [`DispersiveShallowWater.energy_total_modified!`](@ref).
 """
 function energy_total_modified(q_global, equations::AbstractShallowWaterEquations, cache)
     e = similar(q_global.x[begin])
     return energy_total_modified!(e, q_global, equations, cache)
 end
 
+"""
+    energy_total_modified!(e, q_global, equations::AbstractShallowWaterEquations, cache)
+
+In-place version of [`energy_total_modified`](@ref).
+"""
 function energy_total_modified!(e, q_global, equations::AbstractShallowWaterEquations,
                                 cache)
     # `q_global` is an `ArrayPartition` of the primitive variables at all nodes
@@ -300,6 +308,11 @@ Alias for [`energy_total_modified`](@ref).
     return entropy_modified!(e, q_global, equations, cache)
 end
 
+"""
+    entropy_modified!(e, q_global, equations::AbstractShallowWaterEquations, cache)
+
+In-place version of [`entropy_modified`](@ref).
+"""
 @inline entropy_modified!(e, q_global, equations, cache) = energy_total_modified!(e,
                                                                                   q_global,
                                                                                   equations,
--- a/src/equations/equations.jl
+++ b/src/equations/equations.jl
@@ -274,13 +274,14 @@ terms are present.
 function energy_total_modified(q_global, equations::AbstractShallowWaterEquations, cache)
     # `q_global` is an `ArrayPartition` of the primitive variables at all nodes
     @assert nvariables(equations) == length(q_global.x)
+    # tmp1 is always in the cache
+    @unpack tmp1 = cache
 
-    e = similar(q_global.x[begin])
     for i in eachindex(q_global.x[begin])
-        e[i] = energy_total(get_node_vars(q_global, equations, i), equations)
+        tmp1[i] = energy_total(get_node_vars(q_global, equations, i), equations)
     end
 
-    return e
+    return tmp1
 end
 
 varnames(::typeof(energy_total_modified), equations) = ("e_modified",)
@@ -458,6 +459,7 @@ function solve_system_matrix!(dv, system_matrix, rhs,
         cholesky!(factorization, system_matrix; check = false)
         if issuccess(factorization)
             scale_by_mass_matrix!(rhs, D1)
+            # see https://github.com/JoshuaLampert/DispersiveShallowWater.jl/issues/122
             dv .= factorization \ rhs
         else
             # The factorization may fail if the time step is too large
--- a/src/equations/equations.jl
+++ b/src/equations/equations.jl
@@ -272,16 +272,19 @@ contributions.
 terms are present.
 """
 function energy_total_modified(q_global, equations::AbstractShallowWaterEquations, cache)
+    e = similar(q_global.x[begin])
+    return energy_total_modified!(e, q_global, equations, cache)
+end
+
+function energy_total_modified!(e, q_global, equations::AbstractShallowWaterEquations, cache)
     # `q_global` is an `ArrayPartition` of the primitive variables at all nodes
     @assert nvariables(equations) == length(q_global.x)
-    # tmp1 is always in the cache
-    @unpack tmp1 = cache
 
     for i in eachindex(q_global.x[begin])
-        tmp1[i] = energy_total(get_node_vars(q_global, equations, i), equations)
+        e[i] = energy_total(get_node_vars(q_global, equations, i), equations)
     end
 
-    return tmp1
+    return e
 end
 
 varnames(::typeof(energy_total_modified), equations) = ("e_modified",)
@@ -294,6 +297,8 @@ Alias for [`energy_total_modified`](@ref).
 @inline function entropy_modified(q_global, equations::AbstractShallowWaterEquations, cache)
     energy_total_modified(q_global, equations, cache)
 end
+
+@inline entropy_modified!(e, q_global, equations, cache) = energy_total_modified!(e, q_global, equations, cache)
 
 varnames(::typeof(entropy_modified), equations) = ("U_modified",)
 
--- a/src/equations/equations.jl
+++ b/src/equations/equations.jl
@@ -189,7 +189,7 @@ of the correct length `nvariables(equations)`.
 function bathymetry end
 
 """
-        lake_at_rest_error(q, equations::AbstractShallowWaterEquations)
+    lake_at_rest_error(q, equations::AbstractShallowWaterEquations)
 
 Calculate the error for the "lake-at-rest" test case where the
 [`waterheight_total`](@ref) ``\\eta = h + b`` should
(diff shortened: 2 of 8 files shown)


src/equations/svaerd_kalisch_1d.jl
```diff
@@ -174,129 +182,188 @@ function create_cache(mesh, equations::SvaerdKalischEquations1D,
...
-            dv[:] = factorization \ tmp1
-        else
-            fill!(dv, NaN)
-        end
+    @trixi_timeit timer() "assemble system matrix" begin
+        system_matrix = assemble_system_matrix!(cache, h, D1, D1_central, equations)
+    end
+    @trixi_timeit timer() "solving elliptic system" begin
+        tmp1 .= dv
+        solve_system_matrix!(dv, system_matrix, tmp1,
```
JoshuaLampert, line 358: We cannot set `tmp1` in the hyperbolic part immediately because we need `dv` already to be changed for the source terms. I also tried to directly pass `dv` also as third argument to `solve_system_matrix!`, which also works, but seems a bit dangerous to me. Do you think it's save to pass `dv` directly or would you keep it like this or do you have any other idea?

JoshuaLampert:
I plan to make a similar PR for the BBM-BBM equations in the next days, but this will be breaking because it will unify the two equation systems.

src/equations/svaerd_kalisch_1d.jl
ranocha: It's better to keep it like this. There might be subtle issues assuming that `tmp1` and `dv` do not alias when we switch to `ldiv!`.

src/equations/svaerd_kalisch_1d.jl
```diff
@@ -1,8 +1,7 @@
 @doc raw"""
-    SvaerdKalischEquations1D(; gravity_constant, eta0 = 0.0,
-                               alpha = 0.0,
-                               beta = 0.2308939393939394,
-                               gamma = 0.04034343434343434)
+    SvaerdKalischEquations1D(bathymetry_type = bathymetry_variable;
```
ranocha, line 2: Maybe add a note that currently only `bathymetry_variable` is supported? Adding new capabilities will not be breaking.

JoshuaLampert: Done in ce8f281f6f8a08c0cb4d4c5c254ae1a8c9a85aca.

src/equations/equations.jl
```diff
@@ -274,13 +274,14 @@ terms are present.
...
+    @unpack tmp1 = cache
 
-    e = similar(q_global.x[begin])
     for i in eachindex(q_global.x[begin])
-        e[i] = energy_total(get_node_vars(q_global, equations, i), equations)
+        tmp1[i] = energy_total(get_node_vars(q_global, equations, i), equations)
     end
 
-    return e
+    return tmp1
```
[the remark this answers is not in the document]
JoshuaLampert: Sounds good.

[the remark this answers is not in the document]
JoshuaLampert: I made a proposal in 5f609fb342e096271ed01694a1ca315f61034a18, which seems to work. However, I am not completely convinced by this approach. I'll comment on some lines, where I am not really happy with.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -846,23 +821,18 @@ For a [`bathymetry_variable`](@ref) the total modified energy has the additional
 `q_global` is a vector of the primitive variables at ALL nodes.
 `cache` needs to hold the SBP operators used by the `solver`.
 """
-function energy_total_modified(q_global,
-                               equations::SerreGreenNaghdiEquations1D,
-                               cache)
+function energy_total_modified!(e, q_global,
```
JoshuaLampert, line 824: I would like to only need to define the `energy_total_modified!` version for the different equations, but since this is an implementation detail I would not like to present this to the user. Thus, I would like to only document `energy_total_modified`. Now we have a docstring `energy_total_modified` for the function `energy_total_modified!`, which is even weirder. Do you have a better suggestion to achieve this?

src/semidiscretization.jl
```diff
@@ -147,18 +147,12 @@ end
...
-    quantity = func(q, semi.equations, semi.cache)
-    integrate(quantity, semi)
-end
-
 function integrate_quantity!(quantity,
                              func::Union{typeof(energy_total_modified),
                                          typeof(entropy_modified)}, q,
                              semi::Semidiscretization)
-    integrate_quantity(func, q, semi)
+    energy_total_modified!(quantity, q, semi.equations, semi.cache)
```
JoshuaLampert, line 154: Here, I hardcoce `energy_total_modified!`, which works because `energy_total_modified` and `entropy_modified` are doing the same thing, but which is certainly not nice. I would like to use something like `func!`. Can we achieve this somehow?

src/callbacks_step/analysis.jl
```diff
@@ -336,7 +336,7 @@ end
 # the derivative of the velocity `v_x`.
 function analyze(quantity::Union{typeof(energy_total_modified), typeof(entropy_modified)},
                  q, t, semi::Semidiscretization)
-    integrate_quantity(quantity, q, semi)
+    integrate_quantity!(semi.cache.tmp1, quantity, q, semi)
```
JoshuaLampert, line 339: We mutate `semi.cache.tmp1` here, but `analyze` does not have a `!`, which is necessary to make the dispatch work. Should we leave it like this? We could also call both versions `analyze!` eventhough the other one is non-mutating. Any thoughts, @ranocha?

JoshuaLampert: Now that I think of it: We also allocate a vector for the other cases (other `quantity` than `energy_total_modified`) by calling `integrate_quantity`. We could also call `integrate_quantity!` in those cases and pass `semi.cache.tmp1` as first argument. In that case `analyze` would always mutate and the problem above would be solved.

JoshuaLampert: I think we could even completely get rid of the dispatching in this file by defining
```julia
function analyze!(semi::Semidiscretization, quantity, q, t)
  integrate_quantity!(semi.cache.tmp1, quantity, q, semi)
end
```
and slightly rewriting the ` integrate_quantity!` function for the usual `quantities`, which are not `energy_total_modified` or `entropy_modified`.

JoshuaLampert: I implemented that in 8fec4157ee65ffe67eb8e1a483184de6c4527b1e. This also simplifies the relaxation callback, which is quite nice.

src/equations/equations.jl
```diff
@@ -274,13 +274,14 @@ terms are present.
 function energy_total_modified(q_global, equations::AbstractShallowWaterEquations, cache)
     # `q_global` is an `ArrayPartition` of the primitive variables at all nodes
     @assert nvariables(equations) == length(q_global.x)
+    # tmp1 is always in the cache
+    @unpack tmp1 = cache
```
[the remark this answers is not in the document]
JoshuaLampert: I am not sure, I fully understand. So `get_tmp` would just be `get_tmp(cache) = cache.tmp1` to deliver a temporary storage for one field? Where exactly do we need a temporary storage for a full `ArrayPartition`? The `@unpack tmp1 = semi.cache` in the relaxation callback is also just for one field.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -433,7 +433,7 @@ end
 
 # The entropy/energy takes the whole `q` for every point in space
 """
-    energy_total_modified(q_global, equations::HyperbolicSerreGreenNaghdiEquations1D, cache)
+    energy_total_modified!(q_global, equations::HyperbolicSerreGreenNaghdiEquations1D, cache)
```
ranocha, line 436:
```suggestion
    energy_total_modified!(e, q_global, equations::HyperbolicSerreGreenNaghdiEquations1D, cache)
```

We also need to update the docstring accordingly. We may also consider exporting this function - or should prefix it with `DispersiveShallowWater.`. We may also mention this in the general docstring of `energy_total_modified`.

src/equations/serre_green_naghdi_1d.jl
ranocha: Ah, I see. I don't like this since Documenter.jl will still show that it's associated with `energy_total_modified!`, isn't it? I don't have a better solution than just replicating the general `energy_total_modified` specialized to each equation and attach the docstring to it.

src/semidiscretization.jl
ranocha: Maybe something like 

```julia
inplace_version(::typeof(energy_total_modified)) = energy_total_modified!
inplace_version(::typeof(entropy_modified)) = entropy_modified!
```

JoshuaLampert: Looks good. Ideally, I would not need to implement this for every function. I hacked the following together:

```julia
julia> inplace_version(f) = getfield(@__MODULE__, Symbol(string(nameof(f)) * "!"))

julia> inplace_version(energy_total_modified)
energy_total_modified! (generic function with 4 methods)

julia> inplace_version(entropy_modified)
entropy_modified! (generic function with 1 method)
```
Do you think it's too hacky to put into DispersiveShallowWater.jl?

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: So we see two different options here:

1. Document `energy_total_modified!' properly and mention it in `energy_total_modified` (and maybe export it) (this is basically your suggestion from above)
2. Replicate the general `energy_total_modified` for each equation and attach the docstring to it.

Although I consider this as implementation detail, which I would like to hide from the user I tend to 1. since I really don't like replicating code unnecessarily. Are you fine with that? If so, I think I would not export `energy_total_modified!` (and thus prefix it with DispersiveShallowWater.jl in the docstring).

src/semidiscretization.jl
ranocha: Is it type stable?

src/equations/serre_green_naghdi_1d.jl
ranocha: Both options are okay - please choose which one you prefer. The code duplication would only include a small amount of code (allocating the output), so it's not too bad, I think.

src/equations/equations.jl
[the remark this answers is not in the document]
ranocha: We may use `get_tmp(equations, cache)` or even including the `solver`, too, to have more options for dispatch. It can be kind of annoying to have to fix one variable name in the `cache`. But it works for now, so we don't have to change it right now.

The full vector cache could be used in 
https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/cf20e3f45485714167f85d9f5c1fc70ed1dd3b1f/src/callbacks_step/relaxation.jl#L84

src/semidiscretization.jl
JoshuaLampert: Checking with `@code_warntype` gives:

```julia
julia> inplace_version(f) = getfield(@__MODULE__, Symbol(string(nameof(f)) * "!"))
inplace_version (generic function with 1 method)

julia> foo(x) = 2
foo (generic function with 1 method)

julia> foo!(x) = 3
foo! (generic function with 1 method)

julia> @code_warntype inplace_version(foo)
MethodInstance for inplace_version(::typeof(foo))
  from inplace_version(f) @ Main REPL[1]:1
Arguments
  #self#::Core.Const(inplace_version)
  f::Core.Const(foo)
Body::typeof(foo!)
1 ─ %1 = Main.nameof(f)::Core.Const(:foo)
│   %2 = Main.string(%1)::Core.Const("foo")
│   %3 = (%2 * "!")::Core.Const("foo!")
│   %4 = Main.Symbol(%3)::Core.Const(:foo!)
│   %5 = Main.getfield(Main, %4)::Core.Const(foo!)
└──      return %5
```
The `Core.Const` is fine, isn't it? Or should I check something else?

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: I opted for 1. in [686ae52](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/146/commits/686ae527fe7e5f4ce253668a7c94e47ab1f99eb7).

src/equations/equations.jl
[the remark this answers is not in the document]
JoshuaLampert: Ok, let's keep this for a future PR. I think this one is already big enough.

> The full vector cache could be used in

Ah yes, I forgot about that. That  this is allocating is also already tracked in #11. Thanks!

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -433,7 +433,7 @@ end
 
 # The entropy/energy takes the whole `q` for every point in space
 """
-    energy_total_modified!(q_global, equations::HyperbolicSerreGreenNaghdiEquations1D, cache)
+    DispersiveShallowWater.energy_total_modified!(e, q_global, equations::HyperbolicSerreGreenNaghdiEquations1D, cache)
 
 Return the modified total energy of the primitive variables `q_global` for the
```
ranocha, line 438:
```suggestion
Return the modified total energy `e` of the primitive variables `q_global` for the
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -796,7 +796,7 @@ end
 
 # The modified entropy/energy takes the whole `q` for every point in space
 """
-    energy_total_modified(q_global, equations::SerreGreenNaghdiEquations1D, cache)
+    DispersiveShallowWater.energy_total_modified!(q_global, equations::SerreGreenNaghdiEquations1D, cache)
 
 Return the modified total energy of the primitive variables `q_global` for the
```
ranocha, line 801:
```suggestion
Return the modified total energy `e` of the primitive variables `q_global` for the
```

src/equations/svaerd_kalisch_1d.jl
```diff
@@ -401,7 +401,7 @@ end
 
 # The modified entropy/energy takes the whole `q` for every point in space
 """
-    energy_total_modified(q_global, equations::SvaerdKalischEquations1D, cache)
+    DispersiveShallowWater.energy_total_modified!(q_global, equations::SvaerdKalischEquations1D, cache)
 
 Return the modified total energy of the primitive variables `q_global` for the
```
ranocha, line 406:
```suggestion
    DispersiveShallowWater.energy_total_modified!(e, q_global, equations::SvaerdKalischEquations1D, cache)

Return the modified total energy `e` of the primitive variables `q_global` for the
```

ranocha (changes requested):

src/semidiscretization.jl
ranocha: Yeah, looks good 👍

ranocha:
I only have some minor suggestions for improving the documentation. From my point of view, this PR can be merged afterward. Thanks a lot!

src/semidiscretization.jl
JoshuaLampert: This was not yet implemented. I added `inplace_version` in eb6e3c1a3edce07f10c0ec645a4923d338f16e56.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -819,11 +794,11 @@ end
     return equations.eta0 - D
 end
 
-# The entropy/energy takes the whole `q` for every point in space
+# The modified entropy/energy takes the whole `q` for every point in space
 """
-    energy_total_modified(q_global, equations::SerreGreenNaghdiEquations1D, cache)
+    DispersiveShallowWater.energy_total_modified!(q_global, equations::SerreGreenNaghdiEquations1D, cache)
```
JoshuaLampert, line 799:
```suggestion
    DispersiveShallowWater.energy_total_modified!(e, q_global, equations::SerreGreenNaghdiEquations1D, cache)
```

ranocha (approved):
Thanks

PR #147, opened 2024-08-27
Only one BBM-BBM equation struct for flat and variable bathymetry

JoshuaLampert:
This PR merges the two equation structs for the BBM-BBM equations into only one that holds a `bathymetry_type` as we did for the other equations. This also simplifies many things by allowing to define many functions for `AbstractShallowWaterEquations` instead of for all equations separately. As discussed in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/issues/130#issuecomment-2293818956, I also included a check if the bathymetry is constant for a given `bathymetry_flat`.
In addition, the default `bathymetry_type` for Serre-Green-Naghdi equations is set to `bathymetry_variable` (see https://github.com/JoshuaLampert/DispersiveShallowWater.jl/issues/130#issuecomment-2293818956) to be consistent with the other equations.
Finally, I made some changes for more consistency with the other equations and I eliminated some allocations in the hyperbolic parts of the BBM-BBM equations.

TODO:
- [x] Adjust documentation
- [x] Add News entry
- [x] Adjust test tolerances depending on values from CI


test/test_hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -163,26 +163,26 @@ EXAMPLES_DIR = joinpath(examples_dir(), "hyperbolic_serre_green_naghdi_1d")
                                      "hyperbolic_serre_green_naghdi_dingemans.jl"),
                             tspan=(0.0, 1.0),
                             l2=[
-                                0.2462244051802164,
-                                0.8029763168465812,
-                                2.7915052205763788e-15,
-                                0.5551546220119555,
-                                0.24713044468457937,
+                                0.22618134328164993,
```
JoshuaLampert, line 166: These changes (and the ones further down for the SGN equations) are because the Dingemans initial condition was slightly differently defined for the SGN equations.

examples/bbm_bbm_1d/bbm_bbm_1d_basic.jl
```diff
@@ -4,7 +4,8 @@ using DispersiveShallowWater
 ###############################################################################
 # Semidiscretization of the BBM-BBM equations
 
-equations = BBMBBMEquations1D(gravity_constant = 9.81, D = 2.0)
+bathymetry_type = bathymetry_flat # bathymetry_variable
+equations = BBMBBMEquations1D(bathymetry_type, gravity_constant = 9.81)
```
ranocha, line 8: If we always use this pattern, should we make the bathymetry type a keyword argument?

JoshuaLampert: I did that in 72e0e40df700b9ab77e7f51ee01d5d89c91501c1.

JoshuaLampert:
Tests are passing. This is ready for a review @ranocha (sorry for the big PR).

ranocha (commented):
Thanks a lot! I had a first look and will check the BBM files another time

test/test_bbm_bbm_1d.jl
```diff
@@ -11,15 +11,15 @@ EXAMPLES_DIR = joinpath(examples_dir(), "bbm_bbm_1d")
...
                             tspan=(0.0, 1.0),
-                            l2=[0.0032744047423353103 0.0072467844128150435],
-                            linf=[0.0027330692683324997 0.0037734832582039246],
-                            cons_error=[3.61905099297374e-14 5.684341886080802e-14],
-                            change_waterheight=3.61905099297374e-14,
-                            change_velocity=-5.684341886080802e-14,
-                            change_entropy=0.00019552920957721653,
-                            atol_ints=1e-10) # in order to make CI pass
+                            l2=[0.0032743987817220976 0.007246750130235789 0.0],
+                            linf=[0.0027330553477815656 0.0037734958705577526 0.0],
```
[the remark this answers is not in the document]
JoshuaLampert: No, that shouldn't be the reason because the elixir uses `bathymetry_flat`. I am actually not sure why the values changed. I also thought about that briefly, but didn't find a reason why the values should change. The differences are in the order of 1e-8. I thought they are small enough to accept, but I'm not sure. If you have an idea where the difference could come from, I would be interested.
What is also worth mentioning is that the values for the variable bathymetry did not change, compare the [new numbers](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/147/files#diff-9717cb55bc2dcd5ed4467b470a8c9dca30c901e08488cba485dd29f1d18230cbR60) with the [old numbers](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/147/files#diff-7bd90c8f279687056b5c3767e18bd0799a8c78c9b3b8ada1bc10f338221b7228L15).
Another note worth mentioning: I expect the upwind values a few lines further down to change a bit because before the `BBMBBMEquations1D` didn't use the upwind SBP operators, but only the central one.

src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -73,7 +73,7 @@ References for the hyperbolized Serre-Green-Naghdi system can be found in
   [DOI: 10.1007/s10915-021-01429-8](https://doi.org/10.1007/s10915-021-01429-8)
 
 The semidiscretization implemented here conserves
-- the total water mass (integral of h) as a linear invariant
+- the total water mass (integral of `h`) as a linear invariant
```
ranocha, line 76:
```suggestion
- the total water mass (integral of ``h``) as a linear invariant
```

src/equations/bbm_bbm_1d.jl
```diff
@@ -69,7 +98,12 @@ function initial_condition_manufactured(x, t,
                                         mesh)
     eta = exp(t) * cospi(2 * (x - 2 * t))
     v = exp(t / 2) * sinpi(2 * (x - t / 2))
-    return SVector(eta, v)
+    if equations.bathymetry_type isa BathymetryFlat
+        D = 4
+    else # equations.bathymetry_type isa BathymetryVariable
+        D = 5 + 2 * cospi(2 * x)
+    end
```
ranocha, line 105: This is a (minor) type instability in the `if` block. Since we just use `Float64` everywhere, we could fix it via

```suggestion
    if equations.bathymetry_type isa BathymetryFlat
        D = 4.0
    else # equations.bathymetry_type isa BathymetryVariable
        D = 5 + 2 * cospi(2 * x)
    end
```

src/equations/bbm_bbm_1d.jl
```diff
@@ -108,7 +170,12 @@ function initial_condition_manufactured_reflecting(x, t,
                                                    mesh)
     eta = exp(2 * t) * cospi(x)
     v = exp(t) * x * sinpi(x)
-    return SVector(eta, v)
+    if equations.bathymetry_type isa BathymetryFlat
+        D = 4
```
ranocha, line 174:
```suggestion
        D = 4.0
```
see above

src/equations/bbm_bbm_1d.jl
```diff
@@ -170,122 +346,189 @@ function create_cache(mesh, equations::BBMBBMEquations1D,
...
-#   [DOI: 10.4208/cicp.OA-2020-0119](https://doi.org/10.4208/cicp.OA-2020-0119)
+# Discretization that conserves
+# - the total water (integral of `h`) as a linear invariant
+# - the total energy
+# for periodic boundary conditions, see
+# - Joshua Lampert and Hendrik Ranocha (2024)
+#   Structure-Preserving Numerical Methods for Two Nonlinear Systems of Dispersive Wave Equations
+#   [DOI: 10.48550/arXiv.2402.16669](https://doi.org/10.48550/arXiv.2402.16669)
 function rhs!(dq, q, t, mesh, equations::BBMBBMEquations1D, initial_condition,
               ::BoundaryConditionReflecting, source_terms, solver, cache)
```
ranocha, line 484:
```suggestion
# Discretization that conserves
# - the total water (integral of `h`) as a linear invariant
# - the total energy
# for reflecting boundary conditions, see
# - Joshua Lampert and Hendrik Ranocha (2024)
#   Structure-Preserving Numerical Methods for Two Nonlinear Systems of Dispersive Wave Equations
#   [DOI: 10.48550/arXiv.2402.16669](https://doi.org/10.48550/arXiv.2402.16669)
function rhs!(dq, q, t, mesh, equations::BBMBBMEquations1D, initial_condition,
              ::BoundaryConditionReflecting, source_terms, solver, cache)
```

ranocha (changes requested):
Thanks! I just have a few more minor comments

test/test_bbm_bbm_1d.jl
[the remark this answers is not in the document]
ranocha: We use error-based step size control. Since the length of the state vector `q` has changed, the (relative) norms are a bit different. Thus, I would expect to get exactly the same results if we used constant time step sizes. Maybe it's worth testing this once to see whether it's purely an effect of the adaptive time integration method?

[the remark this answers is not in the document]
JoshuaLampert: That makes sense. Let me check this quickly.

ranocha (approved):
Nice 👍

JoshuaLampert:
Thanks for the review @ranocha! Do you have any other breaking changes in mind? If not, I would merge this and release v0.5.0.

ranocha:
Releasing the new version is fine with me :+1:

Issue #95, opened 2024-03-06
Support other models


JoshuaLampert:
It would be nice to support other dispersive wave equations. These include especially the equations in [H. Ranocha, D. Mitsotakis, D. Ketcheson, _A Broad Class of Conservative Numerical Methods for Dispersive Wave Equations_ (2021)](https://arxiv.org/abs/2006.14802) and the corresponding discretizations using SBP operators developed there. These include the following scalar equations with constant bottom:
- [ ] BBM
- [ ] Fornberg-Whitham
- [ ] Camassa-Holm
- [ ] Degasperis-Procesi
- [ ] Holm-Hone

Implementing these semidiscretizations should be pretty straight forward within DispersiveShallowWater.jl. Other interesting models to look at (requiring to develop energy preserving semidiscretizations using SBP operators first) would be the
- [ ]  Serre-Green-Naghdi equations (e.g. [Busto et al., _On High Order ADER Discontinuous Galerkin Schemes for First Order Hyperbolic Reformulations of Nonlinear Dispersive Systems_ (2021)](https://link.springer.com/article/10.1007/s10915-021-01429-8))
- [ ] equations by Escalante and Morales de Luna ([Escalante and Morales de Luna, _A General Non-hydrostatic Hyperbolic Formulation for Boussinesq Dispersive Shallow Flows and Its Numerical Approximation_ (2020)](https://link.springer.com/article/10.1007/s10915-020-01244-7))

JoshuaLampert:
I'm unsure what the best approach would be to add the KdV equation. So far we have implemented the equations in dimensional form, e.g. the BBM equation as

$$
\eta_t + \sqrt{gD}\eta_x + 3/2\sqrt{g/D}\eta\eta_x - 1/6D^2\eta_{xxt} = 0.
$$

In the literature you often find the more abstracted version

$$
u_t + u_x + uu_x - u_{xxt} = 0.
$$

That's fine because these two versions are the same except for constants, which means you can (basically) solve the abstracted form, e.g., by picking $g = D = 1$.
However, for the KdV equation this is a bit different. I would like to support the dimensional form of the equation, such that $\eta$ has the same (physical) meaning across equations in order to be able to compare the solutions of different equations with each other. This means the KdV equation would look like (see also https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150#discussion_r1761209428):

$$
[...]

PR #150, opened 2024-09-10
Add `BBMEquation1D`

JoshuaLampert:
This is a first step towards supporting a few simpler models, cf. #95.


src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,196 @@
+# TODO: Use this nondimensionalized version or use dimensional version including gravity and depth?
```
JoshuaLampert, line 1: I am not sure whether we should implement the nondimensionalized version as here, which is, AFAICT, more commonly used in the literature (e.g., also in "A Broad Class of Conservative Numerical Methods for Dispersive Wave Equations") or whether we should implement a dimensional form containing the gravitational constant and the (flat or maybe in the future even variable) bathymetry. I would prefer this because it is more consistent to the other equations and is essentially more general because this version can be achieved by setting the constants to 1. A downside, however, would be to invest a little bit more time to generalize the semidiscretization and some other functions. What are your thoughts, @ranocha?

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,196 @@
...
+@doc raw"""
+    BBMEquation1D(; bathymetry_type = bathymetry_flat, eta0 = 0.0)
+
+BBM (Benjamin–Bona–Mahony) equation in one spatial dimension. The equations are given by
+```math
+\begin{aligned}
+  \eta_t + \eta\eta_x + \eta_x - \eta_{xxt} &= 0.
```
JoshuaLampert, line 8: I use $\eta$ here instead of $u$ as in the main reference for the BBM equation for consistency with the other equations.

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,196 @@
...
+"""
+    invariant_cubic(q, equations::BBMEquation1D)
+
+Return the cubic invariant of the primitive variables `q` for the
+[`BBMEquation1D`](@ref). The cubic invariant is given by
+```math
+(\\eta + 1)^3.
+```
+"""
+function invariant_cubic(q, equations::BBMEquation1D)
```
JoshuaLampert, line 191: I'm not sure about the name `invariant_cubic`. Do you have another idea, @ranocha?

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,196 @@
...
+```math
+\\frac{1}{2} \\eta^2 + \\frac{1}{2} \\eta_{xx}.
+```
+
+`q_global` is a vector of the primitive variables at ALL nodes.
+`cache` needs to hold the SBP operators used by the `solver`.
+
+See also [`energy_total_modified`](@ref).
+"""
+function energy_total_modified!(e, q_global, equations::BBMEquation1D, cache)
```
JoshuaLampert, line 167: I used the `energy_total_modified` for this function because it uses derivatives of $\eta$. Does this make sense in your eyes? We don't have an `energy` for the `BBMEquation1D` up to now.

src/equations/equations.jl
```diff
@@ -501,7 +522,6 @@ include("svaerd_kalisch_1d.jl")
 # Serre-Green-Naghdi equations
 abstract type AbstractSerreGreenNaghdiEquations{NDIMS, NVARS} <:
               AbstractShallowWaterEquations{NDIMS, NVARS} end
-const AbstractSerreGreenNaghdiEquations1D = AbstractSerreGreenNaghdiEquations{1}
```
JoshuaLampert, line 504: This was not used.

README.md
```diff
@@ -12,6 +12,7 @@
 **DispersiveShallowWater.jl** is a [Julia](https://julialang.org/) package that implements structure-preserving numerical methods for dispersive shallow water models.
 To date, it provides provably conservative, entropy-conserving and well-balanced numerical schemes for some dispersive shallow water models:
 
+* the [BBM equation](https://doi.org/10.4208/cicp.OA-2020-0119),
```
ranocha, line 15:
```suggestion
* the [Benjamin-Bona-Mahony (BBM) equation, also known as regularized long-wave equation](https://doi.org/10.4208/cicp.OA-2020-0119),
```

docs/src/index.md
```diff
@@ -12,6 +12,7 @@
 [**DispersiveShallowWater.jl**](https://github.com/JoshuaLampert/DispersiveShallowWater.jl) is a [Julia](https://julialang.org/) package that implements structure-preserving numerical methods for dispersive shallow water models.
 To date, it provides provably conservative, entropy-conserving and well-balanced numerical schemes for some dispersive shallow water models:
 
+* the [BBM equation](https://doi.org/10.4208/cicp.OA-2020-0119),
```
ranocha, line 15:
```suggestion
* the [Benjamin-Bona-Mahony (BBM) equation, also known as regularized long-wave equation](https://doi.org/10.4208/cicp.OA-2020-0119),
```

src/equations/bbm_1d.jl
ranocha: This is a question I have been thinking about for quite some time - and I agree with you, it would be better to use the dimensional form to be able to compare result for physical setups.

src/equations/bbm_1d.jl
ranocha: 👍

src/equations/bbm_1d.jl
ranocha: For BBM, the quadratic invariant should be $\eta^2 + \eta_x^2$, not $\eta^2 + \eta_{xx}$? In periodic domains, you can write this as $\eta (\eta - \eta_{xx})$, but it has to be quadratic

src/equations/bbm_1d.jl
JoshuaLampert: Ok, do you know of a good reference, which uses the dimensional form of the BBM equation and presents the exact coefficients in front of each term (e.g., I guess for the $\eta_{xxt}$ it would be $\frac{1}{6}D^2$)? I could probably come up with it by my own, but it would be nice to have something to refer.

JoshuaLampert: A related question: What is actually the velocity in the case of the BBM equation? I guess it is somehow (assumed to be) related to the water height by some algebraic relation instead of by a coupled PDE? If so, how does this relation (assumption) look like exactly? It could be nice to make `velocity(q, equations::BBMEquation1D)` work to return this relation given $\eta$. I think I'm missing a bit the (physical) basics/assumptions of these scalar equations.

JoshuaLampert: If we use a dimensional form, it would probably be also good to have the bathymetry part of the unknowns again (i.e. have two unknowns, $\eta$ and $D$) for consistency with the other equations and to be able to generalize it to a variable bathymetry in the future.

JoshuaLampert:
I'm totally confused by this error: https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/10811172112/job/29990016675?pr=150#step:7:8511. Why does it want to convert 3.0 to an Int? And why does this happen all of the sudden (only on Mac)? Any ideas, @ranocha?

src/equations/bbm_1d.jl
ranocha: I am not sure what we could use as a reasonable "energy" of the system - since the energy we have used so far has always been the shallow water energy. But BBM does not have the shallow water system as a subsystem, so I don't know whether there is a canonical "energy" we can use here

src/equations/bbm_1d.jl
ranocha: It's a good question how to call this. Instead of the plain cubic invariant, we could also use the Hamiltonian, see https://doi.org/10.1007/s10543-023-00992-w
Then, it could make sense to implement the two semidiscretizations conserving either the quadratic invariant or the Hamiltonian containing cubic terms

src/equations/bbm_1d.jl
ranocha: I don't have a good reference for the physical setup in mind. Maybe starting with the original paper of BEnjamin, Bona, and Mahony would be a good idea? Or we could ask Dimitrios Mitsotakis maybe?

JoshuaLampert: I'll make a bit of research.

ranocha:
I have no idea - that's very strange indeed

src/equations/bbm_1d.jl
JoshuaLampert: > For BBM, the quadratic invariant should be η 2 + η x 2 , not η 2 + η x x ? In periodic domains, you can write this as η ( η − η x x ) , but it has to be quadratic

You're right. I was too fast. Let me fix that quickly.

JoshuaLampert: > I am not sure what we could use as a reasonable "energy" of the system - since the energy we have used so far has always been the shallow water energy. But BBM does not have the shallow water system as a subsystem, so I don't know whether there is a canonical "energy" we can use here

Ok, so any idea of a good name for this invariant then?

JoshuaLampert: >> For BBM, the quadratic invariant should be η 2 + η x 2 , not η 2 + η x x ? In periodic domains, you can write this as η ( η − η x x ) , but it has to be quadratic

> You're right. I was too fast. Let me fix that quickly.


I just pushed d7f07eaca0722b581aba151a0fb59d7be5357bdc. I hope this fixes this.

src/equations/bbm_1d.jl
JoshuaLampert: Makes sense. I'll do that. When we want to implement both semis, what would be a good way for an interface to switch between the two versions. We could add an argument to the equation to switch between the two semis. This would be easy to implement, but this is question of the discretization and not of the (physical) equation. So I think it doesn't really make sense to put this into the equation. We could also put it into the `solver`, but then how should we store the additional variable? Should it always be in a `solver` argument, also for other equations, where we don't need such a variable? This feels not very elegant. Any ideas?

JoshuaLampert:
> I have no idea - that's very strange indeed

Now CI passes again 🤷‍♂️

src/equations/bbm_1d.jl
JoshuaLampert: > > I am not sure what we could use as a reasonable "energy" of the system - since the energy we have used so far has always been the shallow water energy. But BBM does not have the shallow water system as a subsystem, so I don't know whether there is a canonical "energy" we can use here

> Ok, so any idea of a good name for this invariant then?

Or do you think `energy_total_modified` (and `entropy_modified`) is fine?

ranocha: `energy_total_modified` is fine, I guess. It is in some sense consistent with what we have so far. But we should document this explicitly in the docstrings (of the equations).

JoshuaLampert: Ok, I agree.

ranocha: > Would you recommend implementing eta * (eta - D2 * eta) instead?

Yes 👍 That's the "correct" way to get this invariant

src/equations/bbm_1d.jl
ranocha: Well, the docstrings of the equations we have so far mention the conserved quantities of the semidiscretizations explicitly, so it does indeed make sense to put this choice into the equations - although this looks a bit strange from a physical point of view.

src/equations/bbm_1d.jl
JoshuaLampert: I switched to `eta * (eta - D2 * eta)` in [673a72d](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150/commits/673a72d7cc7463a527082dba94508417672aeddc).

src/equations/bbm_1d.jl
JoshuaLampert: Ok, so let's put it in the equation. It's maybe not ideal, but I can live with it. And if documented well enough it should also not be too confusing for a user.

src/equations/bbm_1d.jl
ranocha: I think you can take most from the KdV equation, e.g., https://de.wikipedia.org/wiki/Korteweg-de-Vries-Gleichung

src/equations/bbm_1d.jl
JoshuaLampert: I'm thinking of how to implement the Hamiltonian. I would like to implement it in the usual form of a function, which only gets a local `q` and the `equations`. This is possible since it doesn't depend on any spatial derivatives in the case of the BBM equation. But is it save to assume that any Hamiltonian function we will implement in the future will not depend on any spatial derivatives? Because if it would, then we would need to use the interface we have for `energy_total_modified`, which gets the global `q_global` and the `cache` holding the SBP operator(s). I am, e.g., thinking of the cubic invariant of the Camassa-Holm equation $\int(u^3 + uu_x^2)$. Would we call this Hamiltonian? If yes, I would opt for already implementing the global variant for the Hamiltonian now and treat it similarly as the modified energy/entropy, such that we don't need to change that in the future. What are your thoughts on that @ranocha?

ranocha: There can be Hamiltonians that depend on derivatives of the solution. So we will need the more general interface

JoshuaLampert: I implemented the Hamiltonian in [577303c](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150/commits/577303c0e80da745d673078dbd29ea905fab1243) and also added a new parameter `split_form` to switch between the two semidiscretizations.

examples/bbm_1d/bbm_1d_hamiltonian_relaxation.jl
```diff
@@ -0,0 +1,43 @@
+using OrdinaryDiffEq
+using DispersiveShallowWater
+
+###############################################################################
+# Semidiscretization of the BBM equation (conserves the quadratic invariant)
```
ranocha, line 5:
```suggestion
# Semidiscretization of the BBM equation (conserves the cubic Hamiltonian)
```

src/equations/bbm_1d.jl
ranocha: 👍

src/equations/bbm_1d.jl
JoshuaLampert: So this means in our notation, the KdV equation reads in dimensional variables:

$$
\eta_t + \sqrt{gD}\eta_x + 3/2\sqrt{g/D}\eta\eta_x + 1/6\sqrt{gD}D^2\eta_{xxx} = 0
$$

For the BBM equation this would mean:

$$
\eta_t + \sqrt{gD}\eta_x + 3/2\sqrt{g/D}\eta\eta_x - 1/6D^2\eta_{xxt} = 0
$$

I also looked this up in [Whitham, _Linear and Nonlinear Waves_ (1974)](https://eclass.uoa.gr/modules/document/file.php/PHYS289/%CE%92%CE%B9%CE%B2%CE%BB%CE%AF%CE%B1/Whitham_Linear_and_Nonlinear_Waves.pdf) and found these two forms of the KdV and BBM equations in equation (13.99) and unnamed equation after (13.99), where $\gamma$ is defined in (13.94), $c_0 = \sqrt{gh_0}$ and $h_0$ is the undisturbed water depth, which we call $D$. I also double-checked that a nondimensionalization with $u = \eta/D$, $\tilde x = x/D$, $\tilde t = t\sqrt{g/D}$ all coefficients cancel and for the BBM equation we get the version, which is implemented now with the additional factors $3/2$ and $1/6$. If you agree @ranocha, I would put $D$ as a second variable into the equations, put an additional parameter `gravity` in the equations, and include the coefficients into the semidiscretizations and in the initial conditions, which are implemented now.

ranocha: Nice 👍 

Why do you want to introduce $D$ as a second variable into the equations? Isn't it assumed to be constant for the equations above?

JoshuaLampert: Yes, $D$ is assumed to be constant. I wanted to include it nervertheless to be consistent with the other equations and to be able to generalize it to a space- and/or time-dependent bathmetry. But since this is probably never happening, we can also put it as a parameter into the equations. I think it's just a bit odd that in this case $D$ is set in the initial condition for the other equations and as parameter for the BBM equation. On the other hand it might be confusing to set it in the initial condition if it has to be constant.

ranocha: Yes, I agree. I don't see a good reason to look for a variable-bathymetry formulation for simple one-way wave models like these. It makes much more sense to include bidirectional waves first before thinking about variable bathymetry. Thus, I would prefer setting the constant in the equations, even if it is not perfectly consistent with the other equations.

JoshuaLampert: Ok, makes sense. Thanks for your insights!

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,241 @@
+@doc raw"""
+    BBMEquation1D(; bathymetry_type = bathymetry_flat,
+                  gravity_constant, D = 1.0, eta0 = 0.0, split_form = true)
```
ranocha, line 3: Since we only support `bathymetry_flat` right now - and I am not sure whether it makes sense to generalize this as discussed earlier - I wonder whether we should drop the `bathymetry_type` here?

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,241 @@
...
+
+BBM (Benjamin–Bona–Mahony) equation in one spatial dimension. The equations are given by
```
ranocha, line 5:
```suggestion
BBM (Benjamin–Bona–Mahony) equation in one spatial dimension.
The equation is given by
```

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,241 @@
...
+The gravitational constant is denoted by `g` and the constant bottom topography (bathymetry) ``b = \eta_0 - D``,
+where ``\eta_0`` is the constant still-water surface and ``D`` the still-water depth. The water height above
+the bathymetry is therefore given by ``h = \eta - \eta_0 + D``. The BBM equation is only implemented for ``\eta_0 = 0``.
+
+Currently, the equations only support a flat bathymetry, see [`bathymetry_flat`](@ref).
+
+The BBM equation is first described in Benjamin, Bona, and Mahony (1972).
+The semidiscretization implemented here is developed in Ranocha, Mitsotakis, and Ketcheson (2020)
+for `split_form = true` and in Linders, Ranocha, and Birken (2023) for `split_form = false`.
+If `split_form` is `true` a split form in the semidiscretization is used, which conserves
```
ranocha, line 22:
```suggestion
If `split_form` is `true`, a split form in the semidiscretization is used, which conserves
```

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,241 @@
...
+- the total water mass (integral of ``h``) as a linear invariant
+- a quadratic invariant (integral of ``1/2\eta(\eta - 1/6D^2\eta_{xx})``), which is called here [`energy_total_modified`](@ref)
```
ranocha, line 24: Maybe mention the other form ($H^1$-like norm) for periodic boundary conditions?

src/equations/bbm_1d.jl
```diff
@@ -0,0 +1,241 @@
...
+        eta_x_upwind = zero(eta2)
+        cache = (; cache..., eta_x_upwind)
+    end
+    return cache
+end
+
+# Discretization that conserves the mass for eta and the modified energy for periodic boundary conditions, see
+# - Hendrik Ranocha, Dimitrios Mitsotakis and David I. Ketcheson (2020)
+#   A Broad Class of Conservative Numerical Methods for Dispersive Wave Equations
+#   [DOI: 10.4208/cicp.OA-2020-0119](https://doi.org/10.4208/cicp.OA-2020-0119)
```
ranocha, line 140: This needs to be updated based since we can have a split form or not.

src/equations/equations.jl
```diff
@@ -290,15 +307,19 @@ varnames(::typeof(entropy), equations) = ("U",)
...
+    energy_total_modified(q_global, equations, cache)
 
 Return the modified total energy of the primitive variables `q_global` for the
 `equations`. This modified total energy is a conserved quantity and can
 contain additional terms compared to the usual [`energy_total`](@ref).
 For example, for the [`SvaerdKalischEquations1D`](@ref) and the
 [`SerreGreenNaghdiEquations1D`](@ref), it contains additional terms
 depending on the derivative of the velocity ``v_x`` modeling non-hydrostatic
-contributions.
+contributions. For equations, which are not `AbstractShallowWaterEquations`,
```
ranocha, line 318:
```suggestion
contributions. For equations which are not `AbstractShallowWaterEquations`,
```

src/equations/equations.jl
```diff
@@ -290,15 +307,19 @@ varnames(::typeof(entropy), equations) = ("U",)
...
+the modified total energy does not have to be an extension of the usual
+[`energy_total`](@ref) and does not have to be related to a physical energy.
+However, it is still a conserved quantity and may contain derivatives of the
+solution.
```
ranocha, line 322:
```suggestion
However, it is still a conserved quantity for appropriate boundary conditions
and may contain derivatives of the solution.
```

src/equations/equations.jl
```diff
@@ -353,6 +374,28 @@ In-place version of [`entropy_modified`](@ref).
...
+derivatives of the solution.
+
+`q_global` is a vector of the primitive variables at ALL nodes.
+`cache` needs to hold the SBP operators used by the `solver` if non-hydrostatic
+terms are present.
+
+Internally, this function allocates a vector for the output and
+calls [`DispersiveShallowWater.hamiltonian!`](@ref).
+"""
+function hamiltonian(q_global, equations::AbstractEquations, cache)
```
ranocha, line 392:
```suggestion
Internally, this function allocates a vector for the output and
calls [`DispersiveShallowWater.hamiltonian!`](@ref).

!!! note
    This function is not necessarily implemented for all equations.
    See [`DispersiveShallowWater.hamiltonian!`](@ref) for further details
    of equations supporting it.
"""
function hamiltonian(q_global, equations::AbstractEquations, cache)
```

ranocha (changes requested):
Thanks! Please find below some comments/questions.

src/equations/bbm_1d.jl
JoshuaLampert: [05d3fbd](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150/commits/05d3fbd740d070a5229a0f9d457bfb341353264d)

src/equations/bbm_1d.jl
JoshuaLampert: [05d3fbd](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150/commits/05d3fbd740d070a5229a0f9d457bfb341353264d)

src/equations/bbm_1d.jl
JoshuaLampert: [b40b9b4](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/150/commits/b40b9b4ba8e8c228ca703a7e990b402f5295e16e)

ranocha (approved):
Thanks!

JoshuaLampert:
Sorry, CI seems to be a bit weird.

PR #151, opened 2024-09-10
Reuse `pretty_form_utf`

JoshuaLampert:
I noticed that we define the pretty printing of some functions twice: Once for the [`AnalysisCallback`](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/e33d981c7e5a82305158608d2dd6841b347f3ccf/src/callbacks_step/analysis.jl#L333) and once in visualization.jl (and forgot, e.g. ,`energy_total_modified` in visualization.jl). In visualization.jl we only have access to the Symbol and not the function since this is what is stored in `integrals(analysis_callback)`, but we can translate it to the function using `getfield(@__MODULE__, name)`. However, most errors ($L^2$, $L^\infty$ and the conservation error) don't have a corresponding function (the lake-at-rest error on the other hand has). Thus, for simplicity, I just kept the current approach for the errors and define a pretty form in visualization.jl for the errors (this means it is redefined for the lake-at-rest error), but reuse the pretty form from the `AnalysisCallback` for the integrals. This means `pretty_form_utf` dispatches a bit strangely on functions and Symbols. I am open for a better suggestion, @ranocha.

### DIFF
--- a/src/visualization.jl
+++ b/src/visualization.jl
@@ -160,18 +160,8 @@ end
     PlotDataOverTime(semi => sol, x_value, conversion)
 end
 
-function pretty(name)
-    if name == :waterheight_total
-        return "∫η"
-    elseif name == :velocity
-        return "∫v"
-    elseif name in [:discharge, :momentum]
-        return "∫P"
-    elseif name == :entropy
-        return "∫U"
-    elseif name == :entropy_modified
-        return "∫U_mod"
-    elseif name == :l2_error
+function pretty_form_utf(name)
+    if name == :l2_error
         return "L² error"
     elseif name == :linf_error
         return "L∞ error"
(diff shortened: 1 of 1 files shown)


ranocha (approved):
A bit unusual but ok 👍 Thanks

src/visualization.jl
```diff
@@ -196,9 +186,10 @@ end
 
         for (i, (name, integral)) in enumerate(pairs(ints))
             name in exclude && continue
+            name_function = getfield(@__MODULE__, name)
```
ranocha, line 189: After thinking more about it (or better not thinking about it), I noticed that this can be problematic if you define a function for an integral outside of the DispersiveShallowWater.jl module. Wold it make sense to just store/use the functions here?

JoshuaLampert: Good point. I didn't think about that. What exactly do mean by " just store/use the functions here"?

ranocha: The functions determining the analysis integrals, e.g., the ones that can be passed in `extra_analysis_integrals`

JoshuaLampert: Let's continue the discussion in #152, where I made a suggestion.

PR #152, opened 2024-09-11
Use functions from `analysis_integrals`

JoshuaLampert:
See the discussion in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/151#discussion_r1753177919.
I'm not sure if that's what you had in mind, @ranocha, but this should work, shouldn't it? We know that the functions within `cb.affect!.analysis_integrals` have the same order as the entries of the NamedTuple `ints = integrals(cb)`. So we can just get the functions from there.

TODO:
- [x] Add a test

### DIFF
--- a/src/visualization.jl
+++ b/src/visualization.jl
@@ -186,10 +186,10 @@ end
 
         for (i, (name, integral)) in enumerate(pairs(ints))
             name in exclude && continue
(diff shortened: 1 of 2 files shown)


ranocha (approved):
Yes, that's what I had in mind. Thanks!

PR #153, opened 2024-09-19
Use TestItems.jl

JoshuaLampert:
With the current testing setup it is quite cumbersome to run only certain tests. I usually just commented all the other tests by hand. By using [TestItems.jl](https://github.com/julia-vscode/TestItems.jl) and [TestItemRunner.jl](https://github.com/julia-vscode/TestItemRunner.jl) this is much more convenient because single test items can be run separately in VScode. The usual testing workflow using `]test` still works. Additionally, running tests is much faster this way. We also don't need the `@trixi_testset` macro anymore.

See also https://www.julia-vscode.org/docs/stable/userguide/testitems/.


test/test_bbm_bbm_1d.jl
```diff
@@ -1,305 +1,294 @@
...
-                            tspan=(0.0, 1.0),
-                            l2=[5.625847040853183e-15 2.7315317493074205e-15 0.0],
-                            linf=[5.995204332975845e-15 1.12866211315047e-14 0.0],
-                            cons_error=[8.881784197001252e-16 1.6076981933671723e-15 0.0],
-                            change_waterheight=-8.881784197001252e-16,
-                            change_velocity=-1.6076981933671723e-15,
-                            change_entropy=-3.1086244689504383e-15)
-
-        @test_allocations(semi, sol, allocs=10_000)
+    @test_allocations(semi, sol, allocs=2_000)
```
JoshuaLampert, line 218: I needed to increase the allocations from `1_500` to `2_000` for some reason to make the test pass (the number of allocations was `1_920`). I have no idea why this was necessary. The same holds for the upwind allocations of the next `@testitem`.

ranocha (approved):
Sounds reasonable. I didn't read all the code changes in detail.

test/test_bbm_bbm_1d.jl
ranocha: Interesting. I also have no immediate idea. Could be caused by less effective constant propagation or something like this 🤷

JoshuaLampert: Ok, but I think it's fine. It's only marginal.

PR #154, opened 2024-09-23
Allow Fourier and periodic rational derivative operators for `BBMBBMEquations1D` and `SvaerdKalischEquations1D`

### LINKED ISSUE #132: Allow `FourierDerivativeOperator`s for BBM-BBM equations and Svärd-Kalisch equations

JoshuaLampert:
I added support for `PeriodicRationalDerivativeOperator`s as second derivative operator and `FourierDerivativeOperator`s for the BBM-BBM and Svärd-Kalisch equations. I also improved the error thrown for an unknown operator.
Closes #132.

### DIFF
--- a/test/test_bbm_1d.jl
+++ b/test/test_bbm_1d.jl
@@ -41,6 +41,23 @@ end
     end
 
     @test_allocations(semi, sol, allocs=5_000)
+
+    # test PeriodicRationalDerivativeOperator
+    D1 = periodic_derivative_operator(1, accuracy_order, xmin(mesh), xmax(mesh),
+                                      nnodes(mesh))
+    D2 = D1^2
+    solver = Solver(D1, D2)
+    @test_trixi_include(joinpath(EXAMPLES_DIR, "bbm_1d_basic.jl"),
+                        tspan=(0.0, 100.0),
+                        solver=solver,
+                        l2=[0.013447144236536044],
+                        linf=[0.007184057857459125],
+                        cons_error=[1.5543122344752192e-15],
+                        change_waterheight=-1.5543122344752192e-15,
+                        change_entropy_modified=-8.68871272874383e-7,
+                        change_hamiltonian=-3.697687313897191e-6)
+
+    @test_allocations(semi, sol, allocs=5_000)
 end
 
 @testitem "bbm_1d_basic with split_form = false" setup=[
--- a/test/test_svaerd_kalisch_1d.jl
+++ b/test/test_svaerd_kalisch_1d.jl
@@ -17,7 +17,11 @@ end
     @test_allocations(semi, sol, allocs=90_000)
 end
 
-@testitem "svaerd_kalisch_1d_dingemans" setup=[Setup, SvaerdKalischEquations1D, AdditionalImports] begin
+@testitem "svaerd_kalisch_1d_dingemans" setup=[
+    Setup,
+    SvaerdKalischEquations1D,
+    AdditionalImports,
+] begin
     @test_trixi_include(joinpath(EXAMPLES_DIR,
                                  "svaerd_kalisch_1d_dingemans.jl"),
                         tspan=(0.0, 1.0),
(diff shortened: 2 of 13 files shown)


test/test_svaerd_kalisch_1d.jl
```diff
@@ -46,6 +69,20 @@ end
...
+                                 "svaerd_kalisch_1d_dingemans_fourier.jl"),
+                        tspan=(0.0, 1.0),
+                        l2=[0.22799246923254327 0.7520172891302948 0.0],
+                        linf=[0.03671494177483947 0.12129171577180138 0.0],
+                        cons_error=[8.526512829121202e-14 5.3078570574495334e-5 0.0],
+                        change_waterheight=-8.526512829121202e-14,
+                        change_entropy=-0.0002424441479433881,
+                        change_entropy_modified=-4.00007138523506e-9)
+
+    @test_allocations(semi, sol, allocs=13_000_000)
```
JoshuaLampert, line 83: Is it expected that the Fourier operator allocates so much for the Svärd-Kalisch equations?

test/test_bbm_1d.jl
```diff
@@ -41,6 +41,23 @@ end
     end
 
     @test_allocations(semi, sol, allocs=5_000)
+
+    # test PeriodicRationalDerivativeOperator
+    D1 = periodic_derivative_operator(1, accuracy_order, xmin(mesh), xmax(mesh),
+                                      nnodes(mesh))
+    D2 = D1^2
+    solver = Solver(D1, D2)
```
ranocha, line 49: If we go as far as this, it could make sense to add downstream tests to SummationByPartsOperators.jl. I am not sure whether I have documented that `D1^` will necessarily return a `PeriodicRationalDerivativeOperator`. It should be pretty stable, but we use some parts that are less clearly described as public API.

test/test_svaerd_kalisch_1d.jl
ranocha: It likely creates a lot of dense matrices for the linear solves, doesn't it?

ranocha (approved):
Thanks!

test/test_svaerd_kalisch_1d.jl
JoshuaLampert: Yes, it needs to allocate a lot of dense matrices. I mainly just wanted to point out that here we need a lot more allocations than in the other tests. But there is probably nothing to do against it. If you agree we can just continue, I think.

ranocha: Yes, let's go ahead

test/test_bbm_1d.jl
JoshuaLampert: The same also holds for `FourierPolynomialDerivativeOperator`, doesn't it? As far as I can tell, it's also not clearly documented that `D1^` will return a `FourierPolynomialDerivativeOperator` if `D1` is a `FourierDerivativeOperator`.

ranocha: Well, I have not added any documentation to it, so it is indeed not really public but more like an implementation detail to make `^` etc. work - which is documented

> All of these periodic derivative operators support multiplication and addition such that polynomials and rational functions of them can be represented efficiently

already in the README.md (see https://github.com/ranocha/SummationByPartsOperators.jl?tab=readme-ov-file#periodic-domains). 

But we can go ahead and merge this PR here. We don't need to add downstream tests right now, but it may be nice to keep this in mind for the future.

ranocha: The simpler option would be to add some explicit tests that `^` returns the appropriate types with some comments that DispersiveShallowWater.jl uses this interface.

ranocha: > The same also holds for `FourierPolynomialDerivativeOperator`, doesn't it? As far as I can tell, it's also not clearly documented that `D1^` will return a `FourierPolynomialDerivativeOperator` if `D1` is a `FourierDerivativeOperator`.

Yes

PR #158, opened 2024-10-28
add benchmark script and testing

JoshuaLampert:
I recently came across [AirspeedVelocity.jl](https://github.com/MilesCranmer/AirspeedVelocity.jl) and think it is a nice tool to run benchmarks for changes in a PR. The GitHub Action file is (almost) a copy of https://github.com/SymbolicML/DynamicQuantities.jl/blob/21b7468801c773c5072c6db358f2fddcb8529ff9/.github/workflows/benchmark_pr.yml.
For the benchmark test I included some elixirs, which should cover most of the features.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,78 @@
+name: Benchmark a Pull Request
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  pull-requests: write
+
+jobs:
+    benchmark:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v4
+            - uses: julia-actions/setup-julia@v2
+              with:
+                version: "1.10"
+            - uses: julia-actions/cache@v2
+            - name: Extract Package Name from Project.toml
+              id: extract-package-name
+              run: |
+                PACKAGE_NAME=$(grep "^name" Project.toml | sed 's/^name = "\(.*\)"$/\1/')
+                echo "package_name=$PACKAGE_NAME" >> $GITHUB_OUTPUT
+            - name: Build AirspeedVelocity
+              env:
+                JULIA_NUM_THREADS: 2
+              run: |
+                # Lightweight build step, as sometimes the runner runs out of memory:
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.add(;url="https://github.com/MilesCranmer/AirspeedVelocity.jl.git")'
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.build("AirspeedVelocity")'
+            - name: Add ~/.julia/bin to PATH
+              run: |
+                echo "$HOME/.julia/bin" >> $GITHUB_PATH
+            - name: Run benchmarks
+              run: |
+                echo $PATH
+                ls -l ~/.julia/bin
+                mkdir results
+                benchpkg ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --url=${{ github.event.repository.clone_url }} --bench-on="${{github.event.pull_request.head.sha}}" --output-dir=results/ --tune
+            - name: Create plots from benchmarks
+              run: |
+                mkdir -p plots
+                benchpkgplot ${{ steps.extract-package-name.outputs.package_name }} --rev="${{github.event.repository.default_branch}},${{github.event.pull_request.head.sha}}" --npart=10 --format=png --input-dir=results/ --output-dir=plots/
+            - name: Upload plot as artifact
+              uses: actions/upload-artifact@v4
+              with:
+                name: plots
+                path: plots
+            - name: Create markdown table from benchmarks
+              run: |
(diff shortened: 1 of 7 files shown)


JoshuaLampert:
Looks like some test values need to be adjusted. Tests were also failing locally. I used the values I now got locally.

JoshuaLampert:
I'm not sure why there is the `UndefVarError: sol not defined` error in the new benchmark Action. Looks like some namespace issue. I guess that's related to the fact that AirspeedVelocity.jl creates a `module` internally, but I don't know how to fix that. Locally, it runs fine.

ranocha (commented):
Sounds like a nice idea to try out - if it does not cause too much work/trouble to set it up

JoshuaLampert:
I have set it up in another repo, which worked easily and nicely. But I'm not sure what to do with the error here.

JoshuaLampert:
This seems fine now. The plots can be found [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/11571452188?pr=158).
It's not optimal that all elixirs are run for the whole time span (because `trixi_include` creates namespace issues in combination with AirspeedVelocity.jl), but with a total runtime of less than 10 minutes I think it's still reasonable.

ranocha:
> Seems like `trixi_include` is the problem. With `include` instead it works. But then we cannot set a smaller time span for the ODE.

`trixi_include([mod::Module=Main,] elixir::AbstractString; kwargs...)` uses `Main` as module whereas `include` uses the current module. Can you use `@__MODULE__` as first argument to `trixi_include`?

JoshuaLampert:
> > Seems like trixi_include is the problem. With include instead it works. But then we cannot set a smaller time span for the ODE.

> trixi_include([mod::Module=Main,] elixir::AbstractString; kwargs...) uses Main as module whereas include uses the current module. Can you use @__MODULE__ as first argument to trixi_include?

Ah, I see. Thanks! Let' see.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,78 @@
...
+              id: extract-package-name
+              run: |
+                PACKAGE_NAME=$(grep "^name" Project.toml | sed 's/^name = "\(.*\)"$/\1/')
+                echo "package_name=$PACKAGE_NAME" >> $GITHUB_OUTPUT
+            - name: Build AirspeedVelocity
+              env:
+                JULIA_NUM_THREADS: 2
+              run: |
+                # Lightweight build step, as sometimes the runner runs out of memory:
+                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.add(;url="https://github.com/MilesCranmer/AirspeedVelocity.jl.git")'
```
ranocha, line 31: Is the package registered?

test/test_serre_green_naghdi_1d.jl
ranocha, line 1: Do you know why the changes were required?

ranocha (dismissed):
Looks reasonable to me. Benchmarks on GitHub runners can be quite noisy but let's see 👍

ranocha:
> This seems fine now. The plots can be found [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/11571452188?pr=158). It's not optimal that all elixirs are run for the whole time span (because `trixi_include` creates namespace issues in combination with AirspeedVelocity.jl), but with a total runtime of less than 10 minutes I think it's still reasonable.

Is there some way to display the plots in a PR comment like the table above?

test/test_serre_green_naghdi_1d.jl
JoshuaLampert: No, unfortunately I don't. Last time CI ran, it succeeded also using [Julia v1.10.5](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/11016947219/job/30594502637#step:3:10).

JoshuaLampert:
> > This seems fine now. The plots can be found [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/11571452188?pr=158). It's not optimal that all elixirs are run for the whole time span (because trixi_include creates namespace issues in combination with AirspeedVelocity.jl), but with a total runtime of less than 10 minutes I think it's still reasonable.

> Is there some way to display the plots in a PR comment like the table above?

Good question. I agree that this would be nice. Let me see if I can wangle that.

test/test_serre_green_naghdi_1d.jl
ranocha: Weird 🤷

.github/workflows/benchmark.yml
ranocha:
```suggestion
                julia -e 'ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0; import Pkg; Pkg.add("AirspeedVelocity")'
```

JoshuaLampert:
> > > This seems fine now. The plots can be found [here](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/11571452188?pr=158). It's not optimal that all elixirs are run for the whole time span (because trixi_include creates namespace issues in combination with AirspeedVelocity.jl), but with a total runtime of less than 10 minutes I think it's still reasonable.
> >
> > Is there some way to display the plots in a PR comment like the table above?
>
> Good question. I agree that this would be nice. Let me see if I can wangle that.

I think it's not so easy. We would need to upload it somewhere else and then put the link into the comment, cf. https://github.com/peter-evans/create-or-update-comment/issues/68. IMHO, that's a bit too sophisticated. The numbers in the table also show most relevant information.

ranocha (approved):
I see. Thanks!

PR #162, opened 2024-12-02
Use default `arch` in CI

JoshuaLampert:
There are warnings (e.g., https://github.com/JoshuaLampert/DispersiveShallowWater.jl/actions/runs/12108289323) in the CI runs because we request x64 on a macOS runner with arm64. I propose to simply use the default architecture, i.e. to remove the `arch` argument. For the `flag-name` of the coveralls action I followed the example under https://github.com/coverallsapp/github-action?tab=readme-ov-file#complete-parallel-job-example.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,7 +42,7 @@ concurrency:
 jobs:
   test:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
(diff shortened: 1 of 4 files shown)


JoshuaLampert:
So it seems CI uses another Mac machine now, which allocates a bit more for one test and needs some adjusted tolerances.

ranocha (approved):
Thanks

PR #163, opened 2024-12-02
Use OrdinaryDiffEqTsit5.jl and OrdinaryDiffEqLowStorageRK.jl

JoshuaLampert:
Since we only use `Tsit5` and `RDPK3SpFSAL35` in the examples, I thought it would be nice to also only use OrdinaryDiffEqTsit5.jl and OrdinaryDiffEqLowStorageRK.jl in the examples and recommend installing OrdinaryDiffEqTsit5.jl in the README and docs. This should significantly reduce precompile time and generally make things smoother and more lightweight. What do you think, @ranocha?


ranocha:
Sounds good to me! However, we need to wait for Trixi.jl to upgrade SciMLBase.jl etc. for this to work properly, don't we?

JoshuaLampert:
> Sounds good to me! However, we need to wait for Trixi.jl to upgrade SciMLBase.jl etc. for this to work properly, don't we?

We only need that if we want to have Trixi.jl and DispersiveShallowWater.jl in the same project. But this currently fails already with the newest version of DispersiveShallowWater.jl, i.e. installing Trixi.jl in a project, where DispersiveShallowWater.jl is installed will downgrade DispersiveShallowWater.jl to an old version (something like v3... if I remember correctly) because the newer DispersiveShallowWater.jl version require versions of RecursiveArrayTools.jl, which are not compatible with Trixi.jl, cf. https://github.com/trixi-framework/Trixi.jl/pull/2150.

JoshuaLampert:
However, we would need to stop supporting Julia v1.9, I guess.

ranocha:
That's fine with me for this repo 👍

JoshuaLampert:
Ok, this required more version bumps than expected. The SciML repos have pretty high compat bounds sometimes... But Downgrade is finally happy. A couple of remarks:

- The failing benchmark action is expected because it uses the main branch as reference, but the benchmark/Project.toml  from this PR. This means it still wants to use OrdinaryDiffEq.jl, but it cannot because it's not in the new Project.toml. IMHO, this can be ignored in this PR. Benchmark should not be influenced by this PR.

- Codecov is failing probably because of some change in Julia v1.10 compared to v1.9 regarding the "lispy" tuple iteration we use in the `AnalysisCallback`, which means the final call to the empty tuple is not covered anymore. I already saw in previous PRs that these lines are only caught by the Julia v1.9 action, which is disabled now. But I think that's also fine.

- Finally, as a replacement for the Julia v1.9 action I added a test using Julia v1.11. This currently fails for an allocation fix. Looks like Julia v1.11 allocates more for this example. Are you fine simply updating the allocations threshold, @ranocha or do you have an idea where this could come from?

ranocha:
> Ok, this required more version bumps than expected. The SciML repos have pretty high compat bounds sometimes... But Downgrade is finally happy. A couple of remarks:
> 
> * The failing benchmark action is expected because it uses the main branch as reference, but the benchmark/Project.toml  from this PR. This means it still wants to use OrdinaryDiffEq.jl, but it cannot because it's not in the new Project.toml. IMHO, this can be ignored in this PR. Benchmark should not be influenced by this PR.

👍 

> * Codecov is failing probably because of some change in Julia v1.10 compared to v1.9 regarding the "lispy" tuple iteration we use in the `AnalysisCallback`, which means the final call to the empty tuple is not covered anymore. I already saw in previous PRs that these lines are only caught by the Julia v1.9 action, which is disabled now. But I think that's also fine.

👍 

> * Finally, as a replacement for the Julia v1.9 action I added a test using Julia v1.11. This currently fails for an allocation fix. Looks like Julia v1.11 allocates more for this example. Are you fine simply updating the allocations threshold, @ranocha or do you have an idea where this could come from?

That's fine as well. Julia v1.11 introduced the `Memory` type so that array operations can lead to slightly more allocations tracked by Julia - but are often more efficient since the compiler can optimize the code better.

PR #168, opened 2024-12-30
Add linear dispersion relation

### LINKED ISSUE #131: Include linear dispersion relation for all equations
It would be nice to have a function, which returns the linear dispersion relation for each set of equations.

- BBM-BBM, e.g.: [Lampert, Ranocha, Structure-Preserving Numerical Methods for Two Nonlinear Systems of Dispersive Wave Equations (2024)](https://arxiv.org/pdf/2402.16669)
- Svärd-Kalisch equations: [Svärd, Kalisch, A novel energy-bounded Boussinesq model and a well balanced and stable numerical discretisation (2023)](https://arxiv.org/pdf/2302.09924)
- Serre-Green-Naghdi, hyperbolic approximation: [Favrie, Gavrilyuk, A rapid numerical method for solving Serre–Green–Naghdi equations describing long free surface gravity waves (2017)](https://iopscience.iop.org/article/10.1088/1361-6544/aa712d).

JoshuaLampert:
This adds a new feature to compute the dispersion relation and wave speed of the different equations. It also adds a new page in the docs explaining the general concept of dispersion, which uses the new `LinearDispersionRelation` to compare the dispersive behavior of the different equations.
I have three questions:

- [x] Should we provide built-in support for the full dispersion relation? If yes, I'm not sure about the best option. We could add a dummy equation `EulerEquation1D` as I do in the docs and dispatch on that. This, however, has the disadvantage that it might lead to the confusion that the Euler equations are supported by DispersiveShallowWater.jl.
- [ ] I'm not sure about the hyperbolic Serre-Green-Naghdi equations. There is a dispersion relation given in Favrie and Gavrilyuk, _A rapid numerical method for solving Serre-Green-Naghdi equations describing long free surface gravity waves_ in eq. (19), but I don't know to understand this. Since $\tau = 1/h$, I initially thought, we obtain a classical dispersion relation by $\tau_0 = 1/h_0$, but this doesn't work out dimension-wise. That this doesn't work can also be seen by eq. (18), which differs from the one in the other two papers mentioned in the code. There is an additional third power in the denominator. I think this is related to the fact that the equations are written in Lagrangian form and not in the classical variables, but I'm not sure. Do you know how this dispersion relation would needs to be translated to obtain the classical version, @ranocha?
- [x] I wanted to double-check that the linear dispersion relation for the BBM equation written as $\eta_t + \sqrt{gD}\eta_x + \frac{3}{2}\sqrt{\frac{g}{D}}\eta\eta_x - \frac{1}{6}D^2\eta_{xxt} = 0$ gives indeed the same as the one for the BBM-BBM equations, but I get another factor of $5/2$ in front (linearizing with $\eta = h_0 + h'$ and $D = h_0$ gives $h'_t + 5/2 \sqrt{g h_0} h'_x - 1/6 h_0^2h\_{xxt} = 0$). I haven't thought about that when implementing the BBM equation. I would expect the same dispersion relation as for the BBM-BBM equations. Do you have an idea how to deal with that, @ranocha?

Closes #131.


docs/src/dispersion.md
```diff
@@ -0,0 +1,145 @@
+# Dispersive properties of the equations
+
+The equations implemented in DispersiveShallowWater.jl describe the propagation of waves in a shallow water system.
```
ranocha, line 3:
```suggestion
The equations implemented in [DispersiveShallowWater.jl](https://github.com/JoshuaLampert/DispersiveShallowWater.jl)
describe the propagation of waves in a shallow water system.
```

docs/src/dispersion.md
```diff
@@ -0,0 +1,145 @@
...
+The equations are dispersive, meaning that the phase speed of the waves depends on their wavelength. This is in
+contrast to non-dispersive waves, where all waves travel at the same speed, such as the solution to the shallow
+water equations. In this case the phase speed is constantly given by ``c_0 = \sqrt{g h_0}``, where ``g`` is the
```
ranocha, line 6:
```suggestion
contrast to non-dispersive waves, where all waves travel at the same speed, such as the solution to the shallow
water equations linearized around a state with vanishing velocity and constant water height ``h_0``.
In this case the phase speed is constantly given by ``c_0 = \sqrt{g h_0}``, where ``g`` is the
```

docs/src/dispersion.md
```diff
@@ -0,0 +1,145 @@
...
+
+The linear dispersion relation of a wave equation describes the relationship between the angular frequency
+``\omega`` and the wavenumber ``k`` of a wave. The angular frequency is related to the period ``T`` of the wave
+by ``\omega = 2\pi / T`` and the wavenumber is related to the wavelength ``\lambda`` by ``k = 2\pi / \lambda``.
+The simple case of just one wave traveling with speed ``c`` is a harmonic (or plane wave) solution of the form
+``\eta(x, t) = \mathrm{e}^{\mathrm{i}(k x - \omega t)}`` with ``c = \omega / k``. The linear dispersion relation
+of a wave equation is then derived by substituting this solution into a linearized version of the equation and
+solving for ``\omega`` in terms of ``k``.
+
+Since all the equations implemented in DispersiveShallowWater.jl are approximations to the full wave equations
```
ranocha, line 17:
```suggestion
Since all the equations implemented in [DispersiveShallowWater.jl](https://github.com/JoshuaLampert/DispersiveShallowWater.jl)
are approximations to the full wave equations
```

docs/src/dispersion.md
```diff
@@ -0,0 +1,145 @@
...
+given by the Euler equations, their dispersion relations do not describe the exact speed of waves, but only
+an approximation to it, where the approximation usually becomes better for longer wavelengths. The dispersion
+relation of the full wave system is given by
+
+```math
+\omega(k) = \pm \sqrt{g k \tanh(h_0 k)}.
+```
+
+In DispersiveShallowWater.jl, we can investigate the dispersion relations of the different equations. Let us
```
ranocha, line 26:
```suggestion
In [DispersiveShallowWater.jl](https://github.com/JoshuaLampert/DispersiveShallowWater.jl),
we can investigate the dispersion relations of the different equations. Let us
```

docs/src/dispersion.md
```diff
@@ -0,0 +1,145 @@
...
+
+h0 = eta0 = 0.8
+g = 9.81
+disp_rel = LinearDispersionRelation(h0)
+k = 0.01:0.01:5.0
+
+struct EulerEquations1D <: DispersiveShallowWater.AbstractEquations{1, 0}
+    gravity::Float64
+    eta0::Float64
+end
```
ranocha, line 42: I would be fine adding this if we document that it is used only to compute the dispersion relation.

src/dispersion_relation.jl
```diff
@@ -0,0 +1,93 @@
...
+function Base.show(io::IO, disp_rel::LinearDispersionRelation)
+    print(io, "LinearDispersionRelation(h0 = ", disp_rel.ref_height, ")")
+end
+
+Base.broadcastable(disp_rel::LinearDispersionRelation) = Ref(disp_rel)
+
+@doc raw"""
+    wave_speed(disp_rel, equations, k; normalize = false)
+
+Compute the wave speed ``c`` for a given wavenumber ``k`` using the linear dispersion relation ``disp_rel``
```
ranocha, line 24:
```suggestion
Compute the wave speed ``c`` for a given wavenumber ``k`` using the 
[`LinearDispersionRelation`](#ref) `disp_rel`
```

src/dispersion_relation.jl
```diff
@@ -0,0 +1,93 @@
+@doc raw"""
+    LinearDispersionRelation(ref_height)
+
+A struct representing a linear dispersion relation ``\omega(k)`` of an equation. The reference
+water height `h0` is given by `ref_height`.
```
ranocha:
```suggestion
water height `h0` is given by `ref_height`. The linear dispersion relation is
used by [`wave_speed`](@ref).
```

src/dispersion_relation.jl
```diff
@@ -0,0 +1,99 @@
...
+"""
+struct LinearDispersionRelation{RealT <: Real}
+    ref_height::RealT
+end
+
+function Base.show(io::IO, disp_rel::LinearDispersionRelation)
+    print(io, "LinearDispersionRelation(h0 = ", disp_rel.ref_height, ")")
+end
+
+Base.broadcastable(disp_rel::LinearDispersionRelation) = Ref(disp_rel)
```
ranocha, line 19:
```suggestion
Base.broadcastable(disp_rel::LinearDispersionRelation) = (disp_rel,)
```

src/dispersion_relation.jl
```diff
@@ -0,0 +1,99 @@
...
+end
+
+# See, e.g., eq. (45) (α = 1) in
+# - Maria Kazolea, Andrea Filippini, Mario Ricchiuto (2023)
+#   Low dispersion finite volume/element discretization of the enhanced Green-Naghdi equations for
+#   wave propagation, breaking and runup on unstructured meshes
+#   [DOI: 10.1016/j.ocemod.2022.102157](https://doi.org/10.1016/j.ocemod.2022.102157)
+#
+# or eq. (53) (β = 0) in
+# - Didier Clamond Denys Dutykh, Dimitrios Mitsotakis (2017)
```
ranocha, line 83:
```suggestion
# - Didier Clamond, Denys Dutykh, Dimitrios Mitsotakis (2017)
```

ranocha:
> I'm not sure about the hyperbolic Serre-Green-Naghdi equations. There is a dispersion relation given in Favrie and Gavrilyuk, A rapid numerical method for solving Serre-Green-Naghdi equations describing long free surface gravity waves in eq. (19), but I don't know to understand this. Since τ = 1 / h, I initially thought, we obtain a classical dispersion relation by τ_0 = 1 / h_0, but this doesn't work out dimension-wise. That this doesn't work can also be seen by eq. (18), which differs from the one in the other two papers mentioned in the code. There is an additional third power in the denominator. I think this is related to the fact that the equations are written in Lagrangian form and not in the classical variables, but I'm not sure. Do you know how this dispersion relation would needs to be translated to obtain the classical version?

Good question... Maybe it would be best to compute the relation from scratch?

ranocha:
> I wanted to double-check that the linear dispersion relation for the BBM equation written as [...] I haven't thought about that when implementing the BBM equation. I would expect the same dispersion relation as for the BBM-BBM equations. Do you have an idea how to deal with that?

Why do you expect the dispersion relations to be the same?

docs/src/dispersion.md
JoshuaLampert: I added them in [f22eb9c](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/168/commits/f22eb9cf249a79a41838760bd5c870a54d7de7e6).

src/dispersion_relation.jl
JoshuaLampert: I added a similar sentence in [9a62605](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/168/commits/9a62605527924b2d5b9e975fa80b3f36e78afee4) and [89324d5](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/168/commits/89324d561c996d5c3af3cbd66ad07e7410619c2b).

JoshuaLampert:
> > I wanted to double-check that the linear dispersion relation for the BBM equation written as [...] I haven't thought about that when implementing the BBM equation. I would expect the same dispersion relation as for the BBM-BBM equations. Do you have an idea how to deal with that?
> 
> Why do you expect the dispersion relations to be the same?

I think it would be pretty weird if the phase speed for the BBM equation would be always 5/2 as fast as the phase speed of the BBM-BBM equations. This would also mean that for $k \to 0$, $c$ would not converge to $c_0$, but $to $5/2 c_0$, which would also be strange, I think. But I might have an idea, where this discrepancy comes from. If we linearize the BBM equation around $\eta = 0 + \eta'$, i.e., we set $\eta_0 = 0$ (which is also what we are doing for the BBM-BBM equations), we don't get any linear contribution from the $3/2\sqrt{g/D}\eta\eta_x$ term anymore, i.e., we lose the additional 3/2 part, which in the end leads to the same linear dispersion relation. Thus, I think, the BBM equation also only makes sense for $\eta_0$ as for the BBM-BBM equations. To verify this also numerically, I ran two scripts similar to the example in the new section I added in this PR.

<details>
<summary>First version, where we use the true $\eta_0$ in the equations and I changed the linear dispersion relation to include the additional factor 5/2.</summary>

```julia
using DispersiveShallowWater
using Plots
using OrdinaryDiffEqTsit5
using Printf

h0 = eta0 = 0.8
g = 9.81
disp_rel = LinearDispersionRelation(h0)
euler = EulerEquations1D(; gravity_constant = g, eta0 = eta0)
equations = BBMEquation1D(; gravity_constant = g, eta0 = eta0, D = eta0)

wave_number() = 3.0
frequency(k) = disp_rel(euler, k)

function initial_condition_traveling_wave(x, t, equations, mesh)
    k = wave_number()
    omega = frequency(k)
    h0 = equations.eta0
    A = 0.02
    h = A * cos(k * x - omega * t)
    v = sqrt(equations.gravity / k * tanh(k * h0)) * h / h0
    eta = h + h0
    D = equations.eta0
    return SVector(eta)
end

k = wave_number()
coordinates_min = 0
coordinates_max = 10 * pi / k # five waves (wave length = 2pi/k)
N = 512
mesh = Mesh1D(coordinates_min, coordinates_max, N)

# create solver with periodic SBP operators of accuracy order 4
accuracy_order = 4
solver = Solver(mesh, accuracy_order)

# semidiscretization holds all the necessary data structures for the spatial discretization
semi = Semidiscretization(mesh, equations, initial_condition_traveling_wave, solver,
                          boundary_conditions = boundary_condition_periodic)

tspan = (0.0, 3.0)
ode = semidiscretize(semi, tspan)

saveat = range(tspan..., length = 201)
sol = solve(ode, Tsit5(), abstol = 1e-7, reltol = 1e-7,
            save_everystep = false, saveat = saveat, tstops = saveat)

x = 0.2 * (coordinates_max - coordinates_min)
c_euler = wave_speed(disp_rel, euler, k)
c = wave_speed(disp_rel, equations, k)
anim = @animate for step in eachindex(sol.u)
    t = sol.t[step]
    x_t_euler = x + c_euler * t
    x_t = x + c * t
    index = argmin(abs.(DispersiveShallowWater.grid(semi) .- x_t))
    scatter([x_t_euler], [initial_condition_traveling_wave(x_t_euler, t, euler, mesh)],
            color = :blue, label = nothing)
    eta, = sol.u[step].x
    scatter!([x_t], [eta[index]],
             color = :green, label = nothing)
    plot!(semi => sol, plot_initial = true, plot_bathymetry = false,
          conversion = waterheight_total, step = step, legend = :topleft, linewidth = 2,
          plot_title = @sprintf("t = %.3f", t), yrange = (0.77, 0.83),
          linestyles = [:solid :dot], labels = ["Euler" "BBM"],
          color = [:blue :green])
end
gif(anim, "traveling_waves_bbm_nonshift.gif", fps = 25)
```

As a result, we obtain
traveling_waves_bbm_nonshift

We see that this exactly gives the numerical phase speed, which is (of course) much too fast.

</details>

<details>
<summary>Second version, where we use `eta0 = 0.0` in the equations and translate the initial condition and the final solution accordingly (we need the same trick for the BBM-BBM equations to reproduce the phase speed predicted by the linear dispersion relation). Here, I kept the dispersion relation for the BBM equation as the same as for the BBM-BBM equations, i.e., without the factor $5/2$.</summary>


```julia
using DispersiveShallowWater
using Plots
using OrdinaryDiffEqTsit5
using Printf

h0 = eta0 = 0.8
g = 9.81
disp_rel = LinearDispersionRelation(h0)
euler = EulerEquations1D(; gravity_constant = g, eta0 = eta0)
equations = BBMEquation1D(; gravity_constant = g, eta0 = 0.0, D = eta0)
# equations = BBMBBMEquations1D(; gravity_constant = g, eta0 = 0.0)

wave_number() = 3.0
frequency(k) = disp_rel(euler, k)

function initial_condition_traveling_wave(x, t, equations, mesh)
    k = wave_number()
    omega = frequency(k)
    h0 = 0.8
    A = 0.02
    h = A * cos(k * x - omega * t)
    v = sqrt(equations.gravity / k * tanh(k * h0)) * h / h0
    eta = h
    D = h0
    return SVector(eta)
end

k = wave_number()
coordinates_min = 0
coordinates_max = 10 * pi / k # five waves (wave length = 2pi/k)
N = 512
mesh = Mesh1D(coordinates_min, coordinates_max, N)

# create solver with periodic SBP operators of accuracy order 4
accuracy_order = 4
solver = Solver(mesh, accuracy_order)

# semidiscretization holds all the necessary data structures for the spatial discretization
semi = Semidiscretization(mesh, equations, initial_condition_traveling_wave, solver,
                          boundary_conditions = boundary_condition_periodic)

tspan = (0.0, 3.0)
ode = semidiscretize(semi, tspan)

saveat = range(tspan..., length = 201)
sol = solve(ode, Tsit5(), abstol = 1e-7, reltol = 1e-7,
            save_everystep = false, saveat = saveat, tstops = saveat)

x = 0.2 * (coordinates_max - coordinates_min)
c_euler = wave_speed(disp_rel, euler, k)
c = wave_speed(disp_rel, equations, k)
shifted_waterheight(q, equations) = waterheight_total(q, equations) + 0.8
DispersiveShallowWater.varnames(shifted_waterheight, equations) = ("η",)
anim = @animate for step in eachindex(sol.u)
    t = sol.t[step]
    x_t_euler = x + c_euler * t
    x_t = x + c * t
    index = argmin(abs.(DispersiveShallowWater.grid(semi) .- x_t))
    scatter([x_t_euler], [initial_condition_traveling_wave(x_t_euler, t, euler, mesh)],
            color = :blue, label = nothing)
    eta, = sol.u[step].x
    scatter!([x_t], [eta[index] + 0.8],
             color = :green, label = nothing)
    plot!(semi => sol, plot_initial = true, plot_bathymetry = false,
          conversion = shifted_waterheight, step = step, legend = :topleft, linewidth = 2,
          plot_title = @sprintf("t = %.3f", t), yrange = (0.77, 0.83),
          linestyles = [:solid :dot], labels = ["Euler" "BBM"],
          color = [:blue :green])
end
gif(anim, "traveling_waves_bbm_shift.gif", fps = 25)
```

Here, the result is:
traveling_waves_bbm_shift

Again, we match the numerically observed phase speed, but this time the speed makes much more sense as it is only a bit too slow as expected.

</details>

To conclude, I would add a [similar warning as for the BBM-BBM equations](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/0c7c07a425e3a03d0d3ab378c576e11085298ebc/src/equations/bbm_bbm_1d.jl#L62) also for the BBM equation. Under this assumption, both equations have the same dispersion relation.
I am always so confused about these different notations and when we need which assumptions. It seems like no-one is making this explicit, but I think it makes sense now and is consistent.

JoshuaLampert:
> > I'm not sure about the hyperbolic Serre-Green-Naghdi equations. There is a dispersion relation given in Favrie and Gavrilyuk, A rapid numerical method for solving Serre-Green-Naghdi equations describing long free surface gravity waves in eq. (19), but I don't know to understand this. Since τ = 1 / h, I initially thought, we obtain a classical dispersion relation by τ_0 = 1 / h_0, but this doesn't work out dimension-wise. That this doesn't work can also be seen by eq. (18), which differs from the one in the other two papers mentioned in the code. There is an additional third power in the denominator. I think this is related to the fact that the equations are written in Lagrangian form and not in the classical variables, but I'm not sure. Do you know how this dispersion relation would needs to be translated to obtain the classical version?
> 
> Good question... Maybe it would be best to compute the relation from scratch?

I tried to compute the linear dispersion relation for the hyperbolic Serre-Green-Naghdi equations from scratch, but had some issues to eliminate the $\eta$ component (what we call `H` here). I don't want to spend more time on this now. I would need to think about it again. I would be fine with just not defining a linear dispersion relation for the hyperbolic SGN equations for now and merge this PR soon. We can always add the linear dispersion relation later once we found a way to express it. What do you think, @ranocha?

ranocha:
> > > I wanted to double-check that the linear dispersion relation for the BBM equation written as [...] I haven't thought about that when implementing the BBM equation. I would expect the same dispersion relation as for the BBM-BBM equations. Do you have an idea how to deal with that?
> > 
> > 
> > Why do you expect the dispersion relations to be the same?
> 
> I think it would be pretty weird if the phase speed for the BBM equation would be always 5/2 as fast as the phase speed of the BBM-BBM equations. This would also mean that for k → 0 , c would not converge to c 0 , but $to 5 / 2 c 0 , which would also be strange, I think. But I might have an idea, where this discrepancy comes from. If we linearize the BBM equation around η = 0 + η ′ , i.e., we set η 0 = 0 (which is also what we are doing for the BBM-BBM equations), we don't get any linear contribution from the 3 / 2 g / D η η x term anymore, i.e., we lose the additional 3/2 part, which in the end leads to the same linear dispersion relation. Thus, I think, the BBM equation also only makes sense for η 0 as for the BBM-BBM equations. To verify this also numerically, I ran two scripts similar to the example in the new section I added in this PR.
> 
> First version, where we use the true 
> η
> 0
>  in the equations and I changed the linear dispersion relation to include the additional factor 5/2.
> Second version, where we use `eta0 = 0.0` in the equations and translate the initial condition and the final solution accordingly (we need the same trick for the BBM-BBM equations to reproduce the phase speed predicted by the linear dispersion relation). Here, I kept the dispersion relation for the BBM equation as the same as for the BBM-BBM equations, i.e., without the factor 
> 5
> /
> 2
> .
> To conclude, I would add a [similar warning as for the BBM-BBM equations](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/0c7c07a425e3a03d0d3ab378c576e11085298ebc/src/equations/bbm_bbm_1d.jl#L62) also for the BBM equation. Under this assumption, both equations have the same dispersion relation. I am always so confused about these different notations and when we need which assumptions. It seems like no-one is making this explicit, but I think it makes sense now and is consistent.

Nice 👍 I also ran into the issue of needing to set `eta0 = 0` and translating the data accordingly. Can we automate this or generalize the system so that other values are also allowed? (not in this PR of course)

docs/src/dispersion.md
```diff
@@ -0,0 +1,139 @@
...
+contrast to non-dispersive waves, where all waves travel at the same speed, such as the solution to the shallow
+water equations linearized around a state with vanishing velocity and constant water height ``h_0``.
+In this case the phase speed is constantly given by ``c_0 = \sqrt{g h_0}``, where ``g`` is the
+acceleration due to gravity and ``h_0`` is a reference water height.
+
+The linear dispersion relation of a wave equation describes the relationship between the angular frequency
+``\omega`` and the wavenumber ``k`` of a wave. The angular frequency is related to the period ``T`` of the wave
+by ``\omega = 2\pi / T`` and the wavenumber is related to the wavelength ``\lambda`` by ``k = 2\pi / \lambda``.
+The simple case of just one wave traveling with speed ``c`` is a harmonic (or plane wave) solution of the form
+``\eta(x, t) = \mathrm{e}^{\mathrm{i}(k x - \omega t)}`` with ``c = \omega / k``. The linear dispersion relation
```
ranocha, line 15:
```suggestion
``\eta(t, x) = \mathrm{e}^{\mathrm{i}(k x - \omega t)}`` with ``c = \omega / k``. The linear dispersion relation
```

Is there a reason to prefer the order $x, t$ instead of $t, x$? I often use $t$ first since it appears to be more natural when looking at semidiscretizations and writing stuff like $u(t) \in L^2$ etc.

src/dispersion_relation.jl
```diff
@@ -0,0 +1,119 @@
+@doc raw"""
+    LinearDispersionRelation(ref_height)
+
+A struct representing a linear dispersion relation ``\omega(k)`` of an equation. The reference
+water height `h0` is given by `ref_height`. A dispersion relation can be called as
```
ranocha, line 5:
```suggestion
water height ``h_0`` is given by `ref_height`. A dispersion relation can be called as
```

Math mode like in the docstring of `wave_speed`?

JoshuaLampert:
> > > > I wanted to double-check that the linear dispersion relation for the BBM equation written as [...] I haven't thought about that when implementing the BBM equation. I would expect the same dispersion relation as for the BBM-BBM equations. Do you have an idea how to deal with that?
> > > 
> > > 
> > > Why do you expect the dispersion relations to be the same?
> > 
> > 
> > I think it would be pretty weird if the phase speed for the BBM equation would be always 5/2 as fast as the phase speed of the BBM-BBM equations. This would also mean that for k → 0 , c would not converge to c 0 , but $to 5 / 2 c 0 , which would also be strange, I think. But I might have an idea, where this discrepancy comes from. If we linearize the BBM equation around η = 0 + η ′ , i.e., we set η 0 = 0 (which is also what we are doing for the BBM-BBM equations), we don't get any linear contribution from the 3 / 2 g / D η η x term anymore, i.e., we lose the additional 3/2 part, which in the end leads to the same linear dispersion relation. Thus, I think, the BBM equation also only makes sense for η 0 as for the BBM-BBM equations. To verify this also numerically, I ran two scripts similar to the example in the new section I added in this PR.
> > First version, where we use the true
> > η
> > 0
> > in the equations and I changed the linear dispersion relation to include the additional factor 5/2.
> > Second version, where we use `eta0 = 0.0` in the equations and translate the initial condition and the final solution accordingly (we need the same trick for the BBM-BBM equations to reproduce the phase speed predicted by the linear dispersion relation). Here, I kept the dispersion relation for the BBM equation as the same as for the BBM-BBM equations, i.e., without the factor
> > 5
> > /
> > 2
> > .
> > To conclude, I would add a [similar warning as for the BBM-BBM equations](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/0c7c07a425e3a03d0d3ab378c576e11085298ebc/src/equations/bbm_bbm_1d.jl#L62) also for the BBM equation. Under this assumption, both equations have the same dispersion relation. I am always so confused about these different notations and when we need which assumptions. It seems like no-one is making this explicit, but I think it makes sense now and is consistent.
> 
> Nice 👍 I also ran into the issue of needing to set `eta0 = 0` and translating the data accordingly. Can we automate this or generalize the system so that other values are also allowed? (not in this PR of course)

I'm thinking about this again. What we could do for the linear dispersion relation is to set it for the BBM equation as $k\sqrt{g h_0}(1 + 3/2\eta_0/h_0)/(1 + 1/6(h_0k)^2)$. This way we always match the wave speed for any $\eta_0$ and we obtain the classical one by setting $\eta_0 = 0$. I wonder if in principal we want to allow $\eta_0 \neq 0$ or if we always want to explicitly set it 0 for the BBM and the BBM-BBM equations. I would need to think about it again. IIRC, we used $\eta_0 \neq 0$ for the well-balancedness test. I would need to check whether we really need $\eta_0 \neq 0$ there.

src/dispersion_relation.jl
```diff
@@ -0,0 +1,119 @@
...
+`eta0`.
+
+!!! note
+    In DispersiveShallowWater.jl, the Euler equations are *only* used for computing the full linear dispersion
+    relation
+    ```math
+    \omega(k) = \sqrt{g k \tanh(h_0 k)}.
+    ```
+    They cannot be solved as a system of equations.
+"""
```
ranocha, line 56:
```suggestion
!!! note
    In DispersiveShallowWater.jl, the Euler equations are *only* used for computing
    the full linear dispersion relation
    ```math
    \omega(k) = \sqrt{g k \tanh(h_0 k)},
    ```
    see [`LinearDispersionRelation`](@ref) and [`wave_speed`](@ref).
    The `EulerEquations1D` cannot be solved as a system of equations.
"""
```

docs/src/dispersion.md
JoshuaLampert: I think the only reason was that `x` comes before `t` in the initial conditions (same as in Trixi.jl). But I'm fine with using t first.

JoshuaLampert:
I just pushed a commit to generalize the linear dispersions relations for both the BBM and the BBM-BBM equations for general $\eta_0$ and consolidated a more consistent naming of the different variables involved, see [824be7e](https://github.com/JoshuaLampert/DispersiveShallowWater.jl/pull/168/commits/824be7eb4576cb33888149741ed32f7f9da69d12). This way we can use any $\eta_0$ with any equations and we numerically reproduce the wave speed given by the linear dispersion relation without needing to shift anything. I still think that $\eta_0 = 0$ is the best choice in most applications for the BBM and BBM-BBM equations, but I also think that general $\eta_0$ can be supported.

ranocha (approved):

PR #170, opened 2025-01-31
use banded matrix decomposition instead of dense LU for BBM-BBM

ranocha:
sparse matrix times banded matrix constructed a dense matrix


JoshuaLampert (approved):
Thanks!

PR #174, opened 2025-04-16
Rename `gravity_constant` to `gravity`

JoshuaLampert:
To stay consistent with the Trixi.jl ecosystem, cf. https://github.com/trixi-framework/Trixi.jl/pull/2357 and to avoid any confusion.


Project.toml
ranocha, line 1: Remember to update the version number - to "0.6.0-DEV" here or to "0.6.0" when this is merged into `main`.

JoshuaLampert: Yeah, I would directly release v0.6.0 after merging this into `main`. I'm just not sure whether to wait until the Trixi.jl breaking release with this change is out. But I tend to just merge this and release the new version now. Do you have a preference?

ranocha: Releasing the new version is fine with me 👍

PR #180, opened 2025-04-29
Add source terms for SGN with BathymetryFlat

cwittens:
Add source terms for SGN with BathymetryFlat. Will do for BathymetryVariable next. 

I did local testing in `examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl` and everything worked fine. 

I was unsure about `test/test_serre_green_naghdi_1d.jl`. Here I only put in a template, but commented it out, because I was unsure which numerical values to put in. 

Also I am not sure, if any more test are needed, which I missed.

### DIFF
--- a/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
+++ b/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
@@ -19,7 +19,7 @@ coordinates_max = 1.0
 N = 256
 mesh = Mesh1D(coordinates_min, coordinates_max, N)
 
-#= for testing rhs_sgn_upwind! use:
+#= for testing upwind operators use:
 D1 = upwind_operators(periodic_derivative_operator;
                       derivative_order = 1, accuracy_order = 4,
                       xmin = xmin(mesh), xmax = xmax(mesh),
@@ -27,7 +27,7 @@ D1 = upwind_operators(periodic_derivative_operator;
 solver = Solver(D1, nothing) 
 =#
 
-# for testing rhs_sgn_central! use:
+# for testing central operators use:
 acc_order = 4
 solver = Solver(mesh, acc_order)
 
--- a/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
+++ b/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
@@ -1,0 +1,114 @@
+using OrdinaryDiffEqTsit5
+using DispersiveShallowWater
+using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
+
+###############################################################################
+# Semidiscretization of the Serre-Green-Naghdi equations
+
+# or bathymetry_variable instead of bathymetry_flat for later
+equations = SerreGreenNaghdiEquations1D(bathymetry_type = bathymetry_flat,
+                                        gravity = 9.81)
+
+initial_condition = initial_condition_manufactured
+source_terms = source_terms_manufactured
+boundary_conditions = boundary_condition_periodic
+
+# create homogeneous mesh
+coordinates_min = 0.0
+coordinates_max = 1.0
+N = 256
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+
+# create solver with periodic upwind SBP operators of accuracy order 4
+# (note that only central-type with even order are used)
+D1 = upwind_operators(periodic_derivative_operator;
+                      derivative_order = 1, accuracy_order = 4,
+                      xmin = xmin(mesh), xmax = xmax(mesh),
+                      N = nnodes(mesh))
+solver = Solver(D1, nothing)
+
+# semidiscretization holds all the necessary data structures for the spatial discretization
+semi = Semidiscretization(mesh, equations, initial_condition, solver,
+                          boundary_conditions = boundary_conditions,
+                          source_terms = source_terms)
+
+###############################################################################
+# Create `ODEProblem` and run the simulation
+
+tspan = (0.0, 1.5)#(xmax(mesh) - xmin(mesh)) / sqrt(1.2 * equations.gravity)) # one period
+ode = semidiscretize(semi, tspan)
+summary_callback = SummaryCallback()
+analysis_callback = AnalysisCallback(semi; interval = 100,
+                                     extra_analysis_errors = (:conservation_error,),
+                                     extra_analysis_integrals = (waterheight_total,
+                                                                 entropy_modified))
+                                                                 
+callbacks = CallbackSet(analysis_callback, summary_callback)
+callbacks = nothing
+saveat = range(tspan..., length = 50)
+sol = solve(ode, Tsit5(), abstol = 1e-10, reltol = 1e-10,
+            save_everystep = false, callback = callbacks, 
+           # saveat = saveat, 
+            );
+
+# plot(semi => sol)            
+# sol
+
+begin
+    X = DispersiveShallowWater.grid(D1)
+    i = length(sol)
+    t = sol.t[i]
+    h =  sol.u[i].x[1]
+    u =  sol.u[i].x[2]
+    p_h = plot(X, h,  title = "t = $t", xlabel = "", ylabel = "h", label="PDE solution - h")
+    # plot!(p_h, X, h_a,  title = "t = $t", xlabel = "", ylabel = "h", label="Analytical solution")
+    p_u = plot(X, u,  title = "", xlabel = "x", ylabel = "v", label="PDE solution - u")
+    # plot!(p_u, X, u_a,  title = "", xlabel = "x", ylabel = "v", label="Analytical solution")
+    plot(p_h, p_u, layout = (1, 2), size = (700, 500))
+end
+
+maximum(sol.u[end].x[1])
+6.970730622952363
+
+minimum(sol.u[end].x[2])
+0.5540396435522645
+
+
+anim = @animate for i in 1:1:length(sol.t)
+
+
+    X = DispersiveShallowWater.grid(D1)
+    t = sol.t[i]
+    h =  sol.u[i].x[1]
+    u =  sol.u[i].x[2]
+    p_h = plot(X, h,  title = "t = $t", xlabel = "", ylabel = "h", label="PDE solution - h")
+    # plot!(p_h, X, h_a,  title = "t = $t", xlabel = "", ylabel = "h", label="Analytical solution")
+    p_u = plot(X, u,  title = "", xlabel = "x", ylabel = "v", label="PDE solution - u")
+    # plot!(p_u, X, u_a,  title = "", xlabel = "x", ylabel = "v", label="Analytical solution")
+    plot(p_h, p_u, layout = (1, 2), size = (700, 500))
+    
+end
+gif(anim, fps = 10)
+    
+
+
+
+begin # set default plotting options for einheitliche und schöne Plots
+    lw = 2; fontsize = 14; fontsize2 = 18
+
+    default(
+    grid=true, 
+    box=:on,
+    size = (700, 500),
+    dpi = 300,
+    lw = lw,
+    titlefont = font(fontsize2,"Computer Modern"),
+    guidefont = font("Computer Modern"),
+    legendfont = font(fontsize-2,"Computer Modern"),
+    xtickfontsize=fontsize,
+    ytickfontsize=fontsize,
+    ylabelfontsize=fontsize2,
+    xlabelfontsize=fontsize2,
+    )
+end
--- a/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
+++ b/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
@@ -46,10 +46,9 @@ analysis_callback = AnalysisCallback(semi; interval = 10,
                                      extra_analysis_errors = (:conservation_error,),
                                      extra_analysis_integrals = (waterheight_total,
                                                                  entropy_modified))
-                                                                 
+
 callbacks = CallbackSet(analysis_callback, summary_callback)
 
 saveat = range(tspan..., length = 100)
 sol = solve(ode, Tsit5(), abstol = 1e-7, reltol = 1e-7,
             save_everystep = false, callback = callbacks, saveat = saveat)
-            
--- a/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
+++ b/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
@@ -48,7 +48,6 @@ analysis_callback = AnalysisCallback(semi; interval = 10,
                                                                  entropy_modified))
 
 callbacks = CallbackSet(analysis_callback, summary_callback)
-
 saveat = range(tspan..., length = 100)
-sol = solve(ode, Tsit5(), abstol = 1e-7, reltol = 1e-7,
+sol = solve(ode, Tsit5(), abstol = 1e-12, reltol = 1e-12,
             save_everystep = false, callback = callbacks, saveat = saveat)
--- a/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
+++ b/examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
@@ -5,7 +5,7 @@ using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
 ###############################################################################
 # Semidiscretization of the Serre-Green-Naghdi equations
 
-# or bathymetry_variable instead of bathymetry_flat for later
+# for now only bathymetry_flat is implemented (bathymetry_variable soon)
 equations = SerreGreenNaghdiEquations1D(bathymetry_type = bathymetry_flat,
                                         gravity = 9.81)
 
@@ -19,14 +19,17 @@ coordinates_max = 1.0
 N = 256
 mesh = Mesh1D(coordinates_min, coordinates_max, N)
 
-
-# create solver with periodic upwind SBP operators of accuracy order 4
-# (note that only central-type with even order are used)
+#= for testing rhs_sgn_upwind! use:
 D1 = upwind_operators(periodic_derivative_operator;
                       derivative_order = 1, accuracy_order = 4,
                       xmin = xmin(mesh), xmax = xmax(mesh),
                       N = nnodes(mesh))
-solver = Solver(D1, nothing)
+solver = Solver(D1, nothing) 
+=#
+
+# for testing rhs_sgn_central! use:
+acc_order = 4
+solver = Solver(mesh, acc_order)
 
 # semidiscretization holds all the necessary data structures for the spatial discretization
 semi = Semidiscretization(mesh, equations, initial_condition, solver,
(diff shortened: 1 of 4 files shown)


JoshuaLampert:
Thanks a lot! Regarding the tests it's fine to use the values you obtain locally after you verified everything works as expected. You can obtain these values from `errors(analysis_callback)` and `integrals(analysis_callback)`.

JoshuaLampert:
Since the manufactured solution is (obviously) an analytical solution, could you please provide a convergence test to see if we match the expected order of convergence? You can use [`convergence_test`](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/ref/#DispersiveShallowWater.convergence_test-Tuple{Module,%20AbstractString,%20Any}) for that.

JoshuaLampert:
For the formatter you can run:
```
julia -e 'using Pkg; Pkg.add(PackageSpec(name = "JuliaFormatter", version="1.0.60"))'
julia -e 'using JuliaFormatter; format(".")'
```

cwittens:
> julia -e 'using Pkg; Pkg.add(PackageSpec(name = "JuliaFormatter", version="1.0.60"))'
> julia -e 'using JuliaFormatter; format(".")'

I did use (Ctrl-K Ctrl-F) in VSCode now - I think it did the same. https://www.julia-vscode.org/docs/dev/userguide/formatter/

cwittens:
> Since the manufactured solution is (obviously) an analytical solution, could you please provide a convergence test to see if we match the expected order of convergence? You can use [`convergence_test`](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/ref/#DispersiveShallowWater.convergence_test-Tuple%7BModule,%20AbstractString,%20Any%7D) for that.

where? The only reference to convergence_test() I found in `./DispersiveShallowWater.jl/test` was https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/95211460557415ce9793da70a7df5c2d20fc0adc/test/test_unit.jl#L443-463, but this is probably not what you mean.

JoshuaLampert:
> > Thanks a lot! Regarding the tests it's fine to use the values you obtain locally after you verified everything works as expected. You can obtain these values from `errors(analysis_callback)` and `integrals(analysis_callback)`.
> 
> For `@test_allocations(semi, sol, allocs=250_000)` I just use what e.g. @Btime gives me when running sol=(...) and round up? The other values I found, thanks :)

The `allocs` refer to the allocation during one call of `rhs!`, not the whole simulation. What you could do to determine the number is to set it to 0 first, run the test locally (you can directly [do that in VSCode](https://www.julia-vscode.org/docs/stable/userguide/testitems/#Running-test-items-inside-VS-Code)), see the error and how many allocations are reported, and finally round the number up and set it as `allocs`.

JoshuaLampert:
> > julia -e 'using Pkg; Pkg.add(PackageSpec(name = "JuliaFormatter", version="1.0.60"))'
> > julia -e 'using JuliaFormatter; format(".")'
> 
> I did use (Ctrl-K Ctrl-F) in VSCode now - I think it did the same. [julia-vscode.org/docs/dev/userguide/formatter](https://www.julia-vscode.org/docs/dev/userguide/formatter/)
> 
> Edit: ok with all those CI fails, maybe I should start doing format(".") afterall :D

You could use VSCode for that (I would recommend using `(Ctrl-Shift-I)` to format the whole file), but it might be that VSCode uses another version of JuliaFormatter.jl, which might result in another formatting.

cwittens:
Because the tests are failing, was it correct to put in the three components of 
```julia
err = errors(analysis_callback)
err.l2_error[:,end]
3-element Vector{Float64}:
 5.034595145055819e-5
 9.546213964026592e-7
 0.0
```

into `l2=[5.034595145055819e-5 9.546213964026592e-7 0.0],` in `@test_trixi_include(...)`?
Either way, I will try fixing it tomorrow.

JoshuaLampert:
> Because the tests are failing, was it correct to put in the three components of
> 
> ```julia
> err = errors(analysis_callback)
> err.l2_error[:,end]
> 3-element Vector{Float64}:
>  5.034595145055819e-5
>  9.546213964026592e-7
>  0.0
> ```
> 
> into `l2=[5.034595145055819e-5 9.546213964026592e-7 0.0],` in `@test_trixi_include(...)`? Either way, I will try fixing it tomorrow.

Yes, looks correct to me. I don't know why tests are failing. Let's check tomorrow :)

JoshuaLampert:
Looking at the CI logs, the tests only [fail on MacOS](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/actions/runs/14738597696/job/41370825490?pr=180#step:7:16432) with differences of the order of 1e-6, which is a little bit too big to be acceptable in my opinion. This looks to me like the problem is rather ill-posed. Could you please check the condition number of the `system_matrix` to see how ill-posed the linear systems are?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -99,7 +99,69 @@ function initial_condition_convergence_test(x, t, equations::SerreGreenNaghdiEqu
...
+"""
+    initial_condition_manufactured(x, t, equations::SerreGreenNaghdiEquations1D, mesh)
+
+A smooth manufactured solution in combination with [`source_terms_manufactured`](@ref).
+"""
+function initial_condition_manufactured(x, t,
+                                        equations::SerreGreenNaghdiEquations1D,
+                                        mesh)
+    h = 3 + cos(pi * (2 * (x - 2 * t)))
+    v = (1 + sin(pi * (2 * (x - t / 2))))
```
JoshuaLampert, line 114:
```suggestion
    h = 3 + cospi(2 * (x - 2 * t))
    v = (1 + sinpi(2 * (x - t / 2)))
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -422,6 +496,7 @@ function rhs_sgn_upwind!(dq, q, equations, source_terms, cache, ::BathymetryFlat
     # arrays for the water height `h` and the velocity `v`.
     eta, v, D = q.x
     dh, dv, dD = dq.x # dh = deta since b is constant in time
+
```
JoshuaLampert, line 499:
```suggestion
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -478,9 +553,17 @@ function rhs_sgn_upwind!(dq, q, equations, source_terms, cache, ::BathymetryFlat
...
+    # use dv as temporary storage to calculate source terms
+    fill!(dv, zero(eltype(dv)))
+    @trixi_timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations,
+                                                       solver)
+    # add source terms to the right-hand side to be solved for
+    @.. tmp += dv
+
     #   dv .= (Diagonal(h) - D1mat_plus * Diagonal(1/3 .* h.^3) * D1mat_minus) \ tmp
     # but faster since the symbolic factorization is reused.
+
```
JoshuaLampert, line 566:
```suggestion
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -478,9 +553,17 @@ function rhs_sgn_upwind!(dq, q, equations, source_terms, cache, ::BathymetryFlat
...
-    # The code below is equivalent to
+    # add source term
+    # use dv as temporary storage to calculate source terms
+    fill!(dv, zero(eltype(dv)))
+    @trixi_timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations,
+                                                       solver)
+    # add source terms to the right-hand side to be solved for
+    @.. tmp += dv
+
     #   dv .= (Diagonal(h) - D1mat_plus * Diagonal(1/3 .* h.^3) * D1mat_minus) \ tmp
```
JoshuaLampert, line 564:
```suggestion
    # The code below is equivalent to
    #   dv .= (Diagonal(h) - D1mat_plus * Diagonal(1/3 .* h.^3) * D1mat_minus) \ tmp
```
Looks like this part of the comment was lost.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -495,7 +578,9 @@ function rhs_sgn_upwind!(dq, q, equations, source_terms, cache, ::BathymetryFlat
     return nothing
 end
 
-function rhs_sgn_central!(dq, q, equations, source_terms, cache,
+using SparseArrays: spdiagm
+
```
JoshuaLampert, line 582:
```suggestion
```

examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
```diff
@@ -0,0 +1,54 @@
...
+source_terms = source_terms_manufactured
+boundary_conditions = boundary_condition_periodic
+
+# create homogeneous mesh
+coordinates_min = 0.0
+coordinates_max = 1.0
+N = 256
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+#= for testing rhs_sgn_upwind! use:
```
JoshuaLampert, line 22:
```suggestion
#= for testing upwind operators use:
```
Since the name `rhs_sgn_upwind!` is internal, I wouldn't expose it to the user.

examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured.jl
```diff
@@ -0,0 +1,54 @@
...
+D1 = upwind_operators(periodic_derivative_operator;
+                      derivative_order = 1, accuracy_order = 4,
+                      xmin = xmin(mesh), xmax = xmax(mesh),
+                      N = nnodes(mesh))
+solver = Solver(D1, nothing) 
+=#
+
+# for testing rhs_sgn_central! use:
```
JoshuaLampert, line 30:
```suggestion
# for testing central operators use:
```
See above.

JoshuaLampert (changes requested):
Thanks a lot! I left some comments below. I suggest to add support for source terms with a variable bathymetry in a separate PR.

src/equations/serre_green_naghdi_1d.jl
cwittens: this i also just found. I think this was the cause of the failed tests.

cwittens:
> Thanks a lot! I left some comments below. I suggest to add support for source terms with a variable bathymetry in a separate PR.

yes, this was the idea :)

cwittens:
> Looking at the CI logs, the tests only [fail on MacOS](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/actions/runs/14738597696/job/41370825490?pr=180#step:7:16432) with differences of the order of 1e-6, which is a little bit too big to be acceptable in my opinion. This looks to me like the problem is rather ill-posed. Could you please check the condition number of the `system_matrix` to see how ill-posed the linear systems are?

I will look into it. 
I assume the if a ill conditioned system is the problem, the easiest way to fix it will be a different manufactured solution?

cwittens:
> Since the manufactured solution is (obviously) an analytical solution, could you please provide a convergence test to see if we match the expected order of convergence? You can use [`convergence_test`](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/ref/#DispersiveShallowWater.convergence_test-Tuple%7BModule,%20AbstractString,%20Any%7D) for that.

I did run 
```julia
convergence_test(SGN_MANUF_File, [128, 256, 512, 1024, 2048], tspan=(0.0, 1.0))

l2
128  3.97e-05  -         128  1.01e-05  -         128  0.00e+00  -
256  5.18e-05  -0.39     256  9.90e-07  3.35      256  0.00e+00  NaN
512  5.61e-05  -0.11     512  1.08e-06  -0.12     512  0.00e+00  NaN
1024 1.28e-04  -1.19     1024 1.08e-06  0.00      1024 0.00e+00  NaN
2048 2.50e-04  -0.96     2048 1.19e-06  -0.15     2048 0.00e+00  NaN

mean           -0.66     mean           0.77      mean           NaN
linf
128  2.01e-04  -         128  1.52e-05  -         128  0.00e+00  -
256  2.41e-04  -0.26     256  3.92e-06  1.95      256  0.00e+00  NaN
512  3.84e-04  -0.68     512  5.91e-06  -0.59     512  0.00e+00  NaN
1024 9.55e-04  -1.31     1024 6.96e-06  -0.24     1024 0.00e+00  NaN
2048 2.12e-03  -1.15     2048 6.43e-06  0.12      2048 0.00e+00  NaN

mean           -0.85     mean           0.31      mean           NaN

```
but not sure if this is how it is supposed to be, because nowhere  I did but the analytical solution for reference (?). 
And also I don't think a negativ EOC is a good thing :D

cwittens:
Not sure why it now also fails for windows and linux, but the bad condition number might be a reason.
For `serre_green_naghdi_manufactured.jl` I get `cond(Matrix(system_matrix)) = ~863932`
while for `serre_green_naghdi_soliton.jl` I get `cond(Matrix(system_matrix)) = ~28`

I will see what I can do to improve that :)

JoshuaLampert:
> Not sure why it now also fails for windows and linux, but the bad condition number might be a reason. For `serre_green_naghdi_manufactured.jl` I get `cond(Matrix(system_matrix)) = ~863932` while for `serre_green_naghdi_soliton.jl` I get `cond(Matrix(system_matrix)) = ~28`
> 
> I will see what I can do to improve that :)

Thanks, but I think ~863932 seems not too high to have such an impact. Looks like there is something else going on.

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -392,6 +456,14 @@ function rhs_sgn_central!(dq, q, equations, source_terms, cache, ::BathymetryFla
                     p_x)
     end
 
+    # add source term
+    # use dv as temporary storage to calculate source terms
+    fill!(dv, zero(eltype(dv)))
+    @trixi_timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations,
+                                                       solver)
+    # add source terms to the right-hand side to be solved for
+    @.. tmp += dv
```
JoshuaLampert, line 465: I am not sure, but could you try a similar implementation of the source terms as in
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/95211460557415ce9793da70a7df5c2d20fc0adc/src/equations/svaerd_kalisch_1d.jl#L425-L449
i.e., use `dv` in the last computation of the current `tmp` in the hyperbolic terms and then set `tmp .= dv` directly before you solve the elliptic system? Then you would not need to set `dv` to zero and add `dv` to `tmp`.

cwittens:
> > > Since the manufactured solution is (obviously) an analytical solution, could you please provide a convergence test to see if we match the expected order of convergence? You can use [`convergence_test`](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/ref/#DispersiveShallowWater.convergence_test-Tuple%7BModule,%20AbstractString,%20Any%7D) for that.
> > 
> > 
> > I did run
> > ```julia
> > convergence_test(SGN_MANUF_File, [128, 256, 512, 1024, 2048], tspan=(0.0, 1.0))
> > 
> > ####################################################################################################
> > l2
> > η                        v                        D
> > N    error     EOC       N    error     EOC       N    error     EOC
> > 128  3.97e-05  -         128  1.01e-05  -         128  0.00e+00  -
> > 256  5.18e-05  -0.39     256  9.90e-07  3.35      256  0.00e+00  NaN
> > 512  5.61e-05  -0.11     512  1.08e-06  -0.12     512  0.00e+00  NaN
> > 1024 1.28e-04  -1.19     1024 1.08e-06  0.00      1024 0.00e+00  NaN
> > 2048 2.50e-04  -0.96     2048 1.19e-06  -0.15     2048 0.00e+00  NaN
> > 
> > mean           -0.66     mean           0.77      mean           NaN
> > ----------------------------------------------------------------------------------------------------
> > linf
> > η                        v                        D
> > N    error     EOC       N    error     EOC       N    error     EOC
> > 128  2.01e-04  -         128  1.52e-05  -         128  0.00e+00  -
> > 256  2.41e-04  -0.26     256  3.92e-06  1.95      256  0.00e+00  NaN
> > 512  3.84e-04  -0.68     512  5.91e-06  -0.59     512  0.00e+00  NaN
> > 1024 9.55e-04  -1.31     1024 6.96e-06  -0.24     1024 0.00e+00  NaN
> > 2048 2.12e-03  -1.15     2048 6.43e-06  0.12      2048 0.00e+00  NaN
> > 
> > mean           -0.85     mean           0.31      mean           NaN
> > ----------------------------------------------------------------------------------------------------
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but not sure if this is how it is supposed to be, because nowhere I did but the analytical solution for reference (?). And also I don't think a negativ EOC is a good thing :D
> 
> Thanks for the convergence test. Yes, I agree, it seems like something is not working yet. We would expect an EOC of approximately 4 for `η` and `v`.

Setting `abstol = 1e-13, reltol = 1e-13`, I now get:
```julia
l2
128  1.83e-05  -         128  1.00e-05  -         128  0.00e+00  -
256  1.15e-06  4.00      256  6.28e-07  4.00      256  0.00e+00  NaN
512  7.17e-08  4.00      512  3.93e-08  4.00      512  0.00e+00  NaN
1024 4.48e-09  4.00      1024 2.45e-09  4.00      1024 0.00e+00  NaN
2048 2.92e-10  3.94      2048 1.41e-10  4.12      2048 0.00e+00  NaN

mean           3.98      mean           4.03      mean           NaN
linf
128  4.66e-05  -         128  1.35e-05  -         128  0.00e+00  -
256  2.93e-06  3.99      256  8.33e-07  4.02      256  0.00e+00  NaN
512  1.83e-07  4.00      512  5.21e-08  4.00      512  0.00e+00  NaN
1024 1.21e-08  3.92      1024 3.25e-09  4.00      1024 0.00e+00  NaN
2048 9.50e-10  3.67      2048 2.10e-10  3.95      2048 0.00e+00  NaN

mean           3.90      mean           3.99      mean           NaN
```

Should I adjust the l2, linf, change_waterhight, ... values in the testitem to the values corresponding to sol with `abstol = 1e-13, reltol = 1e-13` and commit again?

JoshuaLampert:
Ahh, that makes sense. The time error was dominating. Lowering the time integration tolerances to focus on the spatial error sounds good. I'm wondering, however, whether they need to be that low. Can you maybe try something more like `1e-9` to `1e-12`?
Thanks for finding this!

src/equations/serre_green_naghdi_1d.jl
cwittens: If I understand you right, wouldn't that also imply that when their are no source terms, one would just set dv = "value", do nothing and then copy `dv` over to `tmp`?

But you had me thinking and I think doing 
```julia
        # Split form for energy conservation:
        @.. dv = -(g * h2_x - g * h * h_x
                    +
                    0.5 * h * v2_x
                    -
                    0.5 * v^2 * h_x
                    +
                    0.5 * hv_x * v
                    -
                    0.5 * h * v * v_x
                    +
                    p_x)
    end

    # add source term
    @trixi_timeit timer() "source terms" calc_sources!(dq, q, t, source_terms, equations,
                                                       solver)
                                                       
    dv, tmp = tmp, dv
```

is a more efficient way, with just swapping the pointers to `dv` and `tmp` at the end.

JoshuaLampert: Yes, I think your solution should also work and if it is more efficient than the `tmp .= dv` solution, that's even better. Could you please commit and push that?

JoshuaLampert: Not 100% sure. If it's not working I would suggest to go with the `tmp .= dv` approach. If I understand correctly, your doubts were this may be less efficient. I think `tmp .= dv` should not be too expensive because it does not need to allocate a new vector. In the current version we need to add `dv` to `tmp`, which is redundant. Therefore and for the sake of consistency with the Svärd-Kalisch equations I would go with that solution if you don't see any other issue with it. If you see a performance loss, I would like to see some benchmarks showing that :)

cwittens: ok this may explain why:

```julia
org1 = ones(10)
org2 = ones(10) .+ 1

function foo!(org1, org2)
    
    org1 , org2 = org2 , org1
    org1 .*= 10
    @show org1
    return nothing
end

foo!(org1, org2);

@show org1;
@show org2;
```
gives the output:
```julia
org1 = [20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0]
org1 = [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
org2 = [20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0, 20.0]
```

cwittens:
> Ahh, that makes sense. The time error was dominating. Lowering the time integration tolerances to focus on the spatial error sounds good. I'm wondering, however, whether they need to be that low. Can you maybe try something more like `1e-9` to `1e-12`? Thanks for finding this!

for `eta` I only get a EOC of ~4 with `1e-12` or stricter - for both central and upwind operators.

JoshuaLampert:
> > Ahh, that makes sense. The time error was dominating. Lowering the time integration tolerances to focus on the spatial error sounds good. I'm wondering, however, whether they need to be that low. Can you maybe try something more like `1e-9` to `1e-12`? Thanks for finding this!
> 
> for `eta` I only get a EOC of ~4 with `1e-12` or stricter - for both central and upwind operators.

Ok, the EOC for smaller N will be closer to ~4 as for N = 2048. If we only see a "clean" EOC of 4 until, say, N = 512 or 1024 that should also be fine. Otherwise, setting the tolerances to `1e-12` would also be fine with me.

cwittens:
> > > Ahh, that makes sense. The time error was dominating. Lowering the time integration tolerances to focus on the spatial error sounds good. I'm wondering, however, whether they need to be that low. Can you maybe try something more like `1e-9` to `1e-12`? Thanks for finding this!
> > 
> > 
> > for `eta` I only get a EOC of ~4 with `1e-12` or stricter - for both central and upwind operators.
> 
> Ok, I assume the EOC for smaller N will be closer to ~4 as for N = 2048? If we only see a "clean" EOC of 4 until, say, N = 512 or 1024 that should also be fine. Otherwise, setting the tolerances to `1e-12` would also be fine with me.

What I just got was with Ns = [128, 256, 512, 1024, ]. 
I am just adjusting the l2, ... values in the testset, do the test locally and if everything passes I will commit again, and hopefully everything will be working then.

test/test_serre_green_naghdi_1d.jl
```diff
@@ -297,3 +297,17 @@ end
...
+@testitem "serre_green_naghdi_manufactured.jl with bathymetry_flat" setup=[
+    Setup,
+    SerreGreenNaghdiEquations1D
+] begin
+    @test_trixi_include(joinpath(EXAMPLES_DIR, "serre_green_naghdi_manufactured.jl"),
+                        l2=[1.1464202670717813e-6 6.278194071026083e-7 0.0],
+                        linf=[2.9306942055384866e-6 8.332131531396669e-7 0.0],
+                        cons_error=[8.881784197001252e-16 5.479218779802508e-7 0.0],
+                        change_waterheight=-8.881784197001252e-16,
+                        change_entropy_modified=-1.9177564922756574e-6,)
```
JoshuaLampert, line 310:
```suggestion
                        change_entropy_modified=-1.9177564922756574e-6,
                        atol=1e-11) # to make CI pass
```
This should fix the final error on macOS.

JoshuaLampert (changes requested):
One final fix for CI until this looks good from my side. Thanks again!
@ranocha, do you also want to take a look?

JoshuaLampert (approved):

ranocha (approved):
Thanks!

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -99,7 +99,69 @@ function initial_condition_convergence_test(x, t, equations::SerreGreenNaghdiEqu
...
+For a chosen h and v, in this case
+h = 3 + cos(pi * (2 * (x - 2 * t)))
+v = (1 + sin(pi * (2 * (x - t / 2))))
+
+...
+
+dh = (h_t + hv_x)
+dv = (h * v_t - 1//3 * Dx(h^3 * v_tx) + 1//2 * g * Dx(h^2) + 1//2 * h * Dx(v^2) + p_x)
+
+and after this some substitutions to clean everything up.
```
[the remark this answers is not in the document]
cwittens: and here also for the case of Bathymetry which is not flat:
```julia
@variables x t
Dt = Differential(t)
Dx = Differential(x)

@variables g p pi


h = 5 + cos(pi * (2 * (x - 2 * t)))
v = (1 + sin(pi * (2 * (x - t // 2))))

d = -(3//2 - cos(pi*(2 * x)))
b =  - d


h_t = Dt(h)
v_t = Dt(v)

v_x = Dx(v)
b_x = Dx(b)

v_xx = Dx(v_x)
v_tx = Dt(v_x)

psi = 1 // 4 * h * v * Dx(b_x * v)
# psi = 0 # for mild slope
p = 1 // 3 * h^3 * v_x^2 - 1 // 3 * h^3 * v * v_xx + 1 // 2 * h^2 * v * Dx(b_x * v)
p_x = Dx(p)

b1 = h*v_t
b2 = - 1 // 3 * Dx(h^3 * v_tx)
b3 = 1 // 2 * Dx(h^2 * b_x * v_t)
b4 = - 1 // 2 * h^2 * b_x * v_tx
b5 = h * b_x^2 * v_t
b6 = 1 // 2 * g * Dx(h^2)
b7 = g * h * b_x
b8 = 1 // 2 * h * Dx(v^2)
b9 = p_x
b10 = 3 // 2 * (p / h) * b_x

s1 = h_t + Dx(h * v)
s2 = b1 + b2 + b3 + b4 + b5 + b6 + b7 + b8 + b9 + b10 + psi * b_x

source1 = simplify(expand_derivatives(s1))
source2 = simplify(expand_derivatives(s2))
```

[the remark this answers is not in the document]
JoshuaLampert: Looks good. Just to double-check: For the mild slope case you included the additional factor of 3/4 in `b5`?

PR #190, opened 2025-05-06
added warning for odd accuracy_order

cwittens:
DispersiveShallowWater subsects a central difference operator in most Semidiscretizations. This however is not the case for an odd accuracy_order. In the case of the SGN equation this can for example give an EOC of ~1.7 for `accuracy_order=5`.

I didnt want to do any breaking changes, which is way I decided for a warning and not an error message. Also because for example for the BBM-BBM eq. the reduction wasnt as dramatical. 

Also I am not 100% dedicated to the current warning message itself, it is just a first draft.

### DIFF
--- a/test/test_unit.jl
+++ b/test/test_unit.jl
@@ -12,6 +12,9 @@ end
 @testitem "Solver" setup=[Setup, AdditionalImports] begin
     mesh = Mesh1D(-1.0, 1.0, 10)
     p = 3
+    @test_logs (:warn,
+                "DispersiveShallowWater.jl is expecting a central difference operator. This is not given for an odd accuracy (got $p) order.\n This can lead to a significant reduction in the order of convergence of the solution.") Solver(mesh,
+                                                                                                                                                                                                                                              p)
     solver = @test_nowarn Solver(mesh, p + 1)
     @test_nowarn print(solver)
     @test_nowarn display(solver)


src/solver.jl
```diff
@@ -41,6 +41,9 @@ Create a solver, where the summation-by-parts (SBP) operators are of order `accu
 associated to the `mesh`.
 """
 function Solver(mesh, accuracy_order)
+    if accuracy_order % 2 != 0
```
ranocha, line 44:
```suggestion
    if isodd(accuracy_order)
```

src/solver.jl
```diff
@@ -41,6 +41,9 @@ Create a solver, where the summation-by-parts (SBP) operators are of order `accu
 associated to the `mesh`.
 """
 function Solver(mesh, accuracy_order)
+    if isodd(accuracy_order)
+        @warn "DispersiveShallowWater is suspecting a central difference operator. This is not given for an odd accuracy order.\n This can lead to significant reduction in the order of convergence of the solution."
```
JoshuaLampert, line 45:
```suggestion
        @warn "DispersiveShallowWater is expecting a central difference operator. This is not given for an odd accuracy order.\n This can lead to significant reduction in the order of convergence of the solution."
```
"suspect" sounds so negative :sweat_smile:

JoshuaLampert (changes requested):
The observed order of convergence is not always the `accuracy_order` given to the solver. It can depend on the boundary conditions, the equations, the parity of the number of grid points and possible more. So we cannot generally guarantee the EOC to be `accuracy_order`, but yes, in general odd `accuracy_order`s tend to lead to a reduction of the order of convergence. So we can at least warn the user.

cwittens:
I found out way the test are failing, gonna commit in a few min.

cwittens:
I also wanted to include a test for the warning for an odd acc_order, but 
```julia
    @test_warn "DispersiveShallowWater.jl is expecting a central difference operator. This is not given for an odd accuracy (got 3) order.\n This can lead to a significant reduction in the order of convergence of the solution." Solver(mesh, p) 
```
did not work.

JoshuaLampert (approved):
Thanks!

ranocha (approved):
Thanks!

PR #196, opened 2025-05-10
change default parameter set for Svärd-Kalisch equations

### LINKED ISSUE #192: Svärd-Kalisch
- Maybe switch the default parameters (alpha, beta, gamma) to the version published in JCP, i.e., alpha = 0 = gamma and beta = 1 / 3? 
- Write a docstring

JoshuaLampert:
This changes the default parameter set to set 5, which is isotropic and the one in the published paper of Svärd and Kalisch. I also added the reference to the published version of their paper.
Closes #192.

### DIFF
--- a/docs/src/dispersion.md
+++ b/docs/src/dispersion.md
@@ -50,7 +50,8 @@ bbm = BBMEquation1D(; gravity = g, eta0 = eta0, D = h0)
 c_bbm = wave_speed.(disp_rel, bbm, k; normalize = true)
 plot!(k, c_bbm, label = "BBM")
 
-sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0)
+sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0,
+                              alpha = 0.0, beta = 0.2308939393939394, gamma = 0.04034343434343434)
 c_sk = wave_speed.(disp_rel, sk, k; normalize = true)
 plot!(k, c_sk, label = "Svärd-Kalisch")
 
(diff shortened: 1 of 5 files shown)


JoshuaLampert:
I still use the other model in the dispersion part of the docs because with the new default parameter set the linear dispersion relation of the Svärd-Kalisch equations is the same as the one from the Serre-Green-Naghdi equations and IMHO it would be a bit odd if both lines in the plot of the dispersion relation would be exactly above each other.

docs/src/dispersion.md
```diff
@@ -50,7 +50,8 @@ bbm = BBMEquation1D(; gravity = g, eta0 = eta0, D = h0)
 c_bbm = wave_speed.(disp_rel, bbm, k; normalize = true)
 plot!(k, c_bbm, label = "BBM")
 
-sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0)
+sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0,
+                              alpha = 0.0, beta = 0.2308939393939394, gamma = 0.04034343434343434)
```
ranocha, line 54:
```suggestion
# Optimized set 4 in the preprint
sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0,
                              alpha = 0.0, beta = 0.2308939393939394, gamma = 0.04034343434343434)
```

ranocha (changes requested):
Thanks! I just have a minor suggestion. Could you please add a NEWS.md entry since this is a breaking change?

JoshuaLampert:
> Thanks! I just have a minor suggestion. Could you please add a NEWS.md entry since this is a breaking change?

Thanks for the reminder of the news entry. I added one in [52bfd99](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/196/commits/52bfd99bc4b83e67aafaa3edd528a986a1a9f7e8).

ranocha (approved):

PR #197, opened 2025-05-14
pulled extracting caches out of timer.

cwittens:
Maybe I am not seeing something, but wouldnt it make more sense to have all the `get_temp` outside of the `@trixi_timeit timer() "hyperbolic terms" begin` timer? This would make it more consisting with the other equations.

### DIFF
--- a/src/equations/hyperbolic_serre_green_naghdi_1d.jl
+++ b/src/equations/hyperbolic_serre_green_naghdi_1d.jl
@@ -257,7 +257,7 @@ function create_cache(mesh, equations::HyperbolicSerreGreenNaghdiEquations1D,
     # 5: eta, v, D, w, H
     N = ForwardDiff.pickchunksize(5 * nnodes(mesh))
     template = ones(RealT, nnodes(mesh))
-    h = DiffCache(template, N)
+    h = DiffCache(zero(template), N)
     b = DiffCache(zero(template), N)
     b_x = DiffCache(zero(template), N)
     H_over_h = DiffCache(zero(template), N)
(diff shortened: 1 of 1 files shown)


src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -303,28 +303,28 @@ function rhs!(dq, q, t, mesh,
     dh, dv, dD, dw, dH = dq.x # dh = deta since b is constant in time
     fill!(dD, zero(eltype(dD)))
 
+    # First, we extract temporary storage from the `cache`.
+    # Since we use `DiffCache` from PreallocationTools.jl, we need to extract the
+    # appropriate arrays using `get_tmp` and need to pass an array with the element
+    # type we want to use, e.g., plain `Float64` or some dual numbers when using AD.
+
```
ranocha, line 310:
```suggestion
```

ranocha (commented):
The SGN equations use the same approach, see
 https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/src/equations/serre_green_naghdi_1d.jl#L503-L505

https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/src/equations/serre_green_naghdi_1d.jl#L592-L594

https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/src/equations/serre_green_naghdi_1d.jl#L683-L687

https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/src/equations/serre_green_naghdi_1d.jl#L796-L801

JoshuaLampert:
I assume extracting the cache is basically for free, right? So it doesn't really make a difference to put it in the timer or outside. One could argue that if the extracted caches are only used in the hyperbolic terms (which is the case here), they should also be within that timer. This is not the case, e.g., for the BBM-BBM equations because there part of the cache (e.g. `etav`, `Dv, tmp1`) are used in the "deta hyperbolic" part and other parts (`v2, tmp2`) are used in the "dv hyperbolic" part.  (You could of course split the extraction into several parts within their respective timer, but that would introduce some unnecessary code lines.) But I don't care much.
If extracting the cache takes a noticeable amount of time we could also consider putting it into its own timer, but I doubt this is the case.

cwittens:
From the fact that the benchmarking results between main and 25e7bca are identical, it is probably negligible. 
Sorry for the confusion.

JoshuaLampert:
Just for clarification: The [benchmarking results](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/197#issuecomment-2880715152) runs benchmark on the whole `rhs!` function, not only the parts of the code that are wrapped in `@trixi_timeit timer()` blocks. These timed blocks are only used for the timing printed by the `SummaryCallback`. So to see if extracting the caches takes some time, you would need to compare the results reported by the `SummaryCallback` at the end of a simulation. However, I still think this is negligible.

Issue #144, opened 2024-08-18
Improve documentation


JoshuaLampert:
Up to now we only have one overview page in the documentation, which tries to cover everything in the package. I think it would be nice to split this up in several parts focusing on different aspects (e.g. pages about more theoretical aspects and more tutorial-like pages). Ideas for tutorials:

- [ ] Basic introductory example (also explain about SBP operators)
- [ ] Compare different equations applied to the Dingemans setup
- [ ] Compare different SBP solvers: Especially upwind vs central (upwind leads to higher resolution)
- [ ] Explain influence of using relaxation (exact preservation of nonlinear invariant, quadratic vs. linear error growth)

In addition, it would be nice to have a page explaining the implementation strategy and the interplay between different building blocks (mesh, equations, solver, semidiscretization, ODE solver, callbacks), maybe similar to the diagram in https://trixi-framework.github.io/Trixi.jl/stable/overview/#overview-semidiscretizations.

JoshuaLampert:
If you have any other ideas, feel free to add or change the above comment.

ranocha:
Sounds nice!

PR #198, opened 2025-05-14
Implementation of the 1D KdV equation. 

cwittens:
This PR is still missing the necessary documentation. But I already wanted to make the PR to check if all tests are passing.

This PR includes:

- changed Solver struct so that it can hold 3 operators (/Matrices...) and not only 2.
- KdV 1D eq
- examples and tests for:
   - Normal KdV eq with Tsit5
   - KdV with an implicit solver (rhs! seems to be good for Autodiff) (added OrdinaryDiffEqRosenbrock for this (not sure if correctly thou))
   - KdV with Fourier SBP
   - KdV with wide stencil
   - KdV with manufactured solution

In the future I want to add:
- [ ] split up ode (IMEX)
- [ ] add nondim2prim + prim2nondim functions to be able to compare results from $$u_t + u u_x + u_{xxx}=0$$ PDE to the current implementation $$\eta_t+\sqrt{g D} \eta_x+3 / 2 \sqrt{g / D} \eta \eta_x+1 / 6 \sqrt{g D} D^2 \eta_{x x x} = 0$$
- [ ] more examples
- [ ] add more conversion_functions, right now only waterheight is working


JoshuaLampert:
Is your plan to add the TODOs you mention above in this PR or in (a) separate PR(s)? I would be fine with both.

cwittens:
> Is your plan to add the TODOs you mention above in this PR or in (a) separate PR(s)? I would be fine with both.

I would do additional PRs after this one. Not sure how many yet.

ranocha:
> Any ideas on the checks? The only thing that comes to my mind is less strict tolerances in the test itself or stricter tolerances in `solve()`

Yes, that's reasonable. We have observed this behavior with error-based step size control earlier.

src/solver.jl
```diff
@@ -40,19 +56,25 @@ end
 Create a solver, where the summation-by-parts (SBP) operators are of order `accuracy_order` and
 associated to the `mesh`.
 """
-function Solver(mesh, accuracy_order)
+function Solver(mesh, accuracy_order; third = false)
```
JoshuaLampert, line 59:
```suggestion
function Solver(mesh, accuracy_order)
```
I think it should be fine to always create the third derivative operator, such that we don't need this additional keyword argument as it is not really user friendly. See also below.

src/solver.jl
```diff
@@ -40,19 +56,25 @@ end
...
         @warn "DispersiveShallowWater.jl is expecting a central difference operator. This is not given for an odd accuracy (got $accuracy_order) order.\n This can lead to a significant reduction in the order of convergence of the solution."
     end
     D1 = periodic_derivative_operator(1, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
     D2 = periodic_derivative_operator(2, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
+    if third
+        D3 = periodic_derivative_operator(3, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
+        @assert real(D1) == real(D3)
+    else
+        D3 = nothing
+    end
```
JoshuaLampert, line 70:
```suggestion
    D3 = periodic_derivative_operator(3, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
```
Creating the third order operator should not be so expensive. So I think it would be fine to create it even if we don't need it. This way the user doesn't need to specify whether they need a third order derivative operator or not. See above.

src/solver.jl
```diff
@@ -63,8 +85,10 @@ Both summation-by-parts operators should be associated with the same grid.
 """
 function Solver(D1::AbstractDerivativeOperator{RealT},
                 D2::Union{AbstractDerivativeOperator{RealT}, AbstractMatrix{RealT},
-                          Nothing}) where {RealT}
-    Solver{RealT, typeof(D1), typeof(D2)}(D1, D2)
+                          Nothing} = nothing,
```
JoshuaLampert, line 88: With the new default of `nothing` for `D2` could you please replace [all occurrences of `solver(D1, nothing)`](https://github.com/search?q=repo%3ANumericalMathematics%2FDispersiveShallowWater.jl%20Solver(D1%2C%20nothing)&type=code) by `solver(D1)`?

examples/kdv_1d/kdv_1d_manufactured.jl
```diff
@@ -0,0 +1,47 @@
...
+
+summary_callback = SummaryCallback()
+analysis_callback = AnalysisCallback(semi; interval = 100,
+                                     extra_analysis_errors = (:conservation_error,),
+                                     extra_analysis_integrals = (waterheight_total,
+                                                                 waterheight))
+callbacks = CallbackSet(analysis_callback, summary_callback)
+saveat = range(tspan..., length = 100)
+
+# The Problem is very stiff. Tsit5() will cause a maxiter exceeded error.
```
JoshuaLampert, line 44:
```suggestion
# The problem is very stiff. `Tsit5()` will cause a `maxiter` exceeded error.
```

examples/kdv_1d/kdv_1d_fourier.jl
```diff
@@ -0,0 +1,41 @@
...
+
+summary_callback = SummaryCallback()
+analysis_callback = AnalysisCallback(semi; interval = 100,
+                                     extra_analysis_errors = (:conservation_error,),
+                                     extra_analysis_integrals = (waterheight_total,
+                                                                 waterheight))
+callbacks = CallbackSet(analysis_callback, summary_callback)
+saveat = range(tspan..., length = 100)
+
+# Fourier SBP operators are not working with ForwardDiff, so use a explicit method.
```
JoshuaLampert, line 38:
```suggestion
# Fourier SBP operators are not working with ForwardDiff.jl, so use an explicit method.
```

examples/kdv_1d/kdv_1d_wide_stencil.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# create homogeneous mesh
+coordinates_min = -50.0
+coordinates_max = 50.0
+N = 512
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 4
+
+# using a wide stencil third derivative operator is significantly faster than using
```
JoshuaLampert, line 19: Is a wide stencil operator `D3 = D1^3` really significantly faster than upwind operators or do you mean a narrow stencil operator `D3` as created below?

examples/kdv_1d/kdv_1d_wide_stencil.jl
```diff
@@ -0,0 +1,41 @@
...
+# the upwind operator for the third derivative for a given N. However there are considerably more oscillations.
+# This can be improved by increasing N, which in fact leads to it being slower again.
+# TODO: This text maybe in the docs somewhere?
```
JoshuaLampert, line 22: Good question. I would prefer putting that into the documentation in a new section with general recommendations, explanations, and comparisons (accuracy and performance wise) between different SBP operator types (wide stencil, narrow stencil, upwind, Fourier, DG, CG). This should then be demonstrated with numerical examples. This is basically point three in #144.

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+
+A traveling-wave solution used for convergence tests in a periodic domain, here for dimensional variables.
+"""
+function initial_condition_convergence_test(x, t, equations::KdVEquation1D, mesh)
+    g = gravity(equations)
+    D = equations.D
+
+    A = 1.0 # amplitude (free parameter)
+    K = 1 / 2 * sqrt(3 * A / D^3)
+    c = (sqrt(g * D) + A * sqrt(g * D) / (2 * D))
```
JoshuaLampert, line 41: Would it also be possible to use `c` (or rather the non-dimensional speed $\tilde c$ in $c = \tilde c\sqrt{g D}$) as free parameter as we do for the [BBM equation](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/src/equations/bbm_1d.jl#L73)? It should be equivalent to set `c` as `c = ... * sqrt(g * D)` and then compute `A` based on `c` by basically inverting the above relation, shouldn't it?

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    x_t = mod(x - c * t - xmin(mesh), xmax(mesh) - xmin(mesh)) + xmin(mesh)
+
+    eta = A * sech(K * x_t)^2
+
```
JoshuaLampert, line 46:
```suggestion
    x_t = mod(x - c * t - xmin(mesh), xmax(mesh) - xmin(mesh)) + xmin(mesh)
    eta = A * sech(K * x_t)^2
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+    η = 1 + exp(-t // 2) * sin(pi*(2 * (x - t // 2)))
+    η_t = (Dt(η))
+    η_x = (Dx(η))
+    η_xxx = (Dx(Dx(Dx(η))))
+
+    source_kdv = η_t + sqrt(g * D) * η_x + 3//2 * sqrt(g / D) * η * η_x + 1//6 * sqrt(g * D) * D^2 * η_xxx
+
+    source_julia_form = simplify(expand_derivatives(source_kdv))
+```
```
JoshuaLampert, line 83: I would prefer putting that in a comment either in the code or as a comment here in the PR because it's an implementation details that shouldn't be faced too prominently to the user.

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+    b1 = exp(-t / 2)
+    c0 = sqrt(g * D)
+    c1 = sqrt(g / D)
+
+    dη = -0.5 * a1 * b1 - pi * a2 * b1 +
+         2pi * a2 * c0 * b1 +
+         3pi * a2 * c1 * (1 + a1 * b1) * b1 -
+         (4 / 3) * D^2 * pi^3 * a2 * c0 * b1
+
+    return SVector(dη)
```
JoshuaLampert, line 100:
```suggestion
    dq1 = -0.5 * a1 * b1 - pi * a2 * b1 +
          2pi * a2 * c0 * b1 +
          3pi * a2 * c1 * (1 + a1 * b1) * b1 -
          (4 / 3) * D^2 * pi^3 * a2 * c0 * b1

    return SVector(dq1)
```
I would call this `dq1` for the following reasons:
- consistency with the other source terms
- Without explicitly mentioning it in the (developer) docs, I basically inherited the [style guide from Trixi.jl](https://trixi-framework.github.io/TrixiDocumentation/stable/styleguide/), which doesn't allow non-ASCII characters in executable code.
- `deta` is used as a variable name for the time derivative of $\eta$, which is not the same as the source term (although I just realize that for the same reason `dq1` is not quite accurate because `q1` is just `eta`, but whatever :sweat_smile:).

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    eta_x = DiffCache(zero(template), N)
+    eta_xxx = DiffCache(zero(template), N)
+
+    if solver.D1 isa PeriodicUpwindOperators
+        D1 = solver.D1.central
+
+        # calculate the third derivative operator using upwind operators
+        D3 = sparse(solver.D1.plus) * sparse(solver.D1.minus) * sparse(solver.D1.central)
+
```
JoshuaLampert, line 130:
```suggestion
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,175 @@
...
+        mul!(eta_x, D1, eta)
+
+        # eta_xxx = D3 * eta
+        mul!(eta_xxx, D3, eta)
+
+        @.. deta = -1.0 * (c_0 * eta_x +
+                    c_1 * eta * eta_x +
+                    c_1 * eta2_x +
+                    1 / 6 * c_0 * DD * eta_xxx)
+    end
```
JoshuaLampert, line 169:
```suggestion

        @.. deta = -1.0 * (c_0 * eta_x +
                    c_1 * (eta * eta_x +
                           eta2_x)
    end
        
    @trixi_timeit timer() "third-order derivatives" begin
        # eta_xxx = D3 * eta
        mul!(eta_xxx, D3, eta)
        
        @.. deta -= 1 / 6 * c_0 * DD * eta_xxx
    end
```
The third order derivative terms are not hyperbolic. So I would put them into a separate timer.

test/test_kdv_1d.jl
```diff
@@ -0,0 +1,63 @@
...
+end
+
+@testitem "kdv_1d_fourier" setup=[Setup, KdVEquation1D] begin
+    @test_trixi_include(joinpath(EXAMPLES_DIR, "kdv_1d_fourier.jl"),
+                        tspan=(0.0, 5.0),
+                        l2=[0.0006248516017854692],
+                        linf=[0.00011293320517307625],
+                        cons_error=[4.440892098500626e-15],
+                        change_waterheight=4.440892098500626e-15,
+                        atol=1e-9) # to make CI pass))
```
JoshuaLampert, line 34:
```suggestion
                        atol=1e-9) # to make CI pass
```

test/test_kdv_1d.jl
```diff
@@ -0,0 +1,63 @@
...
+
+    @test_allocations(semi, sol, allocs=5_000)
+end
+
+err = errors(analysis_callback)
+int = integrals(analysis_callback)
+
```
JoshuaLampert, line 41:
```suggestion
```

test/test_kdv_1d.jl
```diff
@@ -0,0 +1,63 @@
...
+@testitem "kdv_1d_manufactured" setup=[Setup, KdVEquation1D] begin
+    @test_trixi_include(joinpath(EXAMPLES_DIR, "kdv_1d_manufactured.jl"),
+                        tspan=(0.0, 1.0),
+                        l2=[8.19260012441373e-8],
+                        linf=[8.736123491281944e-8],
+                        cons_error=[1.545513450551539e-8],
+                        change_waterheight=-1.548159223041523e-8,
+                        atol=1e-9) # to make CI pass)
```
JoshuaLampert, line 49:
```suggestion
                        atol=1e-9) # to make CI pass
```

JoshuaLampert (changes requested):
Thanks a lot! I left some comments below.
Apart from adding the docstring, can you please
- add a news entry in https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/NEWS.md
- add it to the [README.md](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/README.md#L15)
- add it to the [docs landing page](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/docs/src/index.md#L15)?

Could I ask you again for some convergence results, please?

> And I am not sure I did add OrdinaryDiffEqRosenbrock in the right way (?)

Looks good.

test/test_kdv_1d.jl
JoshuaLampert, line 49:
```suggestion
                        atol=1e-9,
                        atol_ints=1e-10) # to make CI pass)
```
This should fix two of the failing tests.

test/test_kdv_1d.jl
JoshuaLampert, line 34:
```suggestion
                        atol=1e-8) # to make CI pass))
```
This should fix the macOS failing test.

JoshuaLampert (changes requested):
I forgot. Could you please also add some unit tests similar to https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/test/test_unit.jl#L97-L143
(of course only the parts that are already in this PR)

examples/kdv_1d/kdv_1d_wide_stencil.jl
cwittens: I mean compared to the D3= Dm*Dc*Dp. 
Just running it was ~3x faster but there where a lot of oscillations still, which only decreased with bigger N
I think it is because of different (worse?) step size control. 
I havent compared them using fixed dt

JoshuaLampert: So you compared `solver = Solver(mesh, accuracy_order; third = true)` to the upwind operators and the first one was ~3x faster? That makes sense to me, but then using a _narrow_ stencil third derivative operator is significantly faster.
The oscillations also make sense. Usually upwind operators are better in reducing oscillations. I would expect even more oscillations with `D1^3` compared to the `solver = Solver(mesh, accuracy_order; third = true)` case.

src/equations/kdv_1d.jl
cwittens: Calculation of the manufactured solution was done the following way:

```julia
    using Symbolics
    @variables x t 
    Dt = Differential(t)
    Dx = Differential(x)
    @variables g D pi
    η = 1 + exp(-t // 2) * sin(pi*(2 * (x - t // 2)))
    η_t = (Dt(η))
    η_x = (Dx(η))
    η_xxx = (Dx(Dx(Dx(η))))
    source_kdv = η_t + sqrt(g * D) * η_x + 3//2 * sqrt(g / D) * η * η_x + 1//6 * sqrt(g * D) * D^2 * η_xxx
    source_julia_form = simplify(expand_derivatives(source_kdv))
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,158 @@
...
+        @.. eta2 = eta^2
+        # eta2_x = D1 * eta2
+        mul!(eta2_x, D1, eta2)
+
+        # eta_x = D1 * eta
+        mul!(eta_x, D1, eta)
+
+        @.. deta = -1.0 * (c_0 * eta_x +
+                    c_1 * (eta * eta_x +
+                           eta2_x)
```
JoshuaLampert, line 144:
```suggestion
                           eta2_x))
```
Sorry, I missed this parenthesis.

examples/kdv_1d/kdv_1d_wide_stencil.jl
```diff
@@ -0,0 +1,41 @@
...
+accuracy_order = 4
+solver = Solver(mesh, accuracy_order; third = true)
```
JoshuaLampert, line 24:
```suggestion
solver = Solver(mesh, accuracy_order)
```

examples/kdv_1d/kdv_1d_wide_stencil.jl
cwittens: yes.

examples/kdv_1d/kdv_1d_wide_stencil.jl
cwittens: Alright, I put it on my ToDo.

src/equations/kdv_1d.jl
cwittens: I will try. (I hope this wont change all test cases thou...)

examples/kdv_1d/kdv_1d_wide_stencil.jl
JoshuaLampert:
```suggestion
# using a narrow stencil third derivative operator is significantly faster than using
```

examples/kdv_1d/kdv_1d_wide_stencil.jl
JoshuaLampert: Thanks!

src/solver.jl
```diff
@@ -63,8 +85,10 @@ Both summation-by-parts operators should be associated with the same grid.
 """
```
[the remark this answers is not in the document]
cwittens: will do.

cwittens:
> > And I am not sure I did add OrdinaryDiffEqRosenbrock in the right way (?)
> 
> Looks good.

Why are there errors in the Downgrade CI then?

JoshuaLampert:
> > > And I am not sure I did add OrdinaryDiffEqRosenbrock in the right way (?)
> > 
> > 
> > Looks good.
> 
> Why are there errors in the Downgrade CI then?

Ah, we need to bump the version to resolve the unsatisfiable requirements.

src/equations/kdv_1d.jl
JoshuaLampert: @ranocha, do you have a preference whether to use `A` or `c` as free parameter? Are there any conventions? We use `c` for the BBM equation, but Collin and I are unsure, which of the two is more intuitive.

JoshuaLampert:
> > > > > And I am not sure I did add OrdinaryDiffEqRosenbrock in the right way (?)
> > > > 
> > > > 
> > > > Looks good.
> > > 
> > > 
> > > Why are there errors in the Downgrade CI then?
> > 
> > 
> > Ah, we need to bump the version to resolve the unsatisfiable requirements.
> 
> how / where do I do that?

So the compat bound for OrdinaryDiffEqRosenbrock.jl [is set to v1.1](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/198/files#diff-ad1f158d5c6f62a7fe387246b89370948e00b895a395dd451ab5106932fcef62R22). The Downgrade action automatically tests whether the compat bounds in our Project.toml and test/Project.toml are set correctly by setting (downgrading) the compat bound to the lowest supported version (so `OrdinaryDiffEqRosenbrock = "1.1"` becomes `OrdinaryDiffEqRosenbrock = "=1.1"`), which means it tries to run the tests with all versions as the minimal ones that are allowed by our Project.toml files. What the error tells you, is that OrdinaryDiffEqRosenbrock.jl@1.1 is not compatible with the other compat bounds because there are packages with the minimal version allowed by our Project.toml that require certain versions of other packages, which are not allowed by OrdinaryDiffEqRosenbrock.jl. In this case
- OrdinaryDiffEqRosenbrock.jl@1.1 [requires at least LinearSolve.jl@2.32](https://github.com/SciML/OrdinaryDiffEq.jl/blob/7a98e43b63691a14e2e3e80f803fa324bee83ad4/lib/OrdinaryDiffEqRosenbrock/Project.toml#L34)
- LinearSolve.jl >= v2.23 (in particular v2.32) [requires at least BandedMatrices.jl v1.5](https://github.com/SciML/LinearSolve.jl/blob/v2.23.0/Project.toml#L64)
- [We allow BandedMatrices.jl v1](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/Project.toml#L29),which downgrades to `BandedMatrices=1.0` as compat bound

This means that on the one hand we need at least BandedMatrices.jl=1.0 from our compat, but on the other hand BandedMatrices>=1.5 indirectly coming from OrdinaryDiffEqRosenbrock.jl=1.1. There are two fixes for this:
1. Either OrdinaryDiffEqRosenbrock.jl (and therefore LinearSolve.jl) allows lower versions of BandedMatrices.jl (specifically v1.0 in our case). I think this would theoretically be possible (not sure about this specific case, but I encountered similar problems, where I think it should), but the SciML people tend to just set pretty high (often unnecessarily high) compat bounds and don't care this breaks Downgrade actions (while at the same time [recommending Downgrade](https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatibility/110063) (well...and [also not recommending it](https://github.com/SciML/OrdinaryDiffEq.jl/issues/2639#issuecomment-2760801931))).
2. We bump our compat bound of BandedMatrices.jl to v1.5. This is of course the way to do because it's easy to do and doesn't require fighting with the whole SciML ecosystem.

But, yeah, generally Downgrade tends to be very annoying, but IMHO enhances code quality.

README.md
```diff
@@ -12,6 +12,7 @@
 **DispersiveShallowWater.jl** is a [Julia](https://julialang.org/) package that implements structure-preserving numerical methods for dispersive shallow water models.
 To date, it provides provably conservative, entropy-conserving and well-balanced numerical schemes for some dispersive shallow water models:
 
+* the [Korteweg–De Vries (KdV) equation the prototypical example of an integrable PDE](https://doi.org/10.1007/s10915-025-02898-x)
```
JoshuaLampert, line 15:
```suggestion
* the [Korteweg–De Vries (KdV) equation as the prototypical example of an integrable PDE](https://doi.org/10.1007/s10915-025-02898-x),
```

docs/src/index.md
```diff
@@ -12,6 +12,7 @@
 [**DispersiveShallowWater.jl**](https://github.com/NumericalMathematics/DispersiveShallowWater.jl) is a [Julia](https://julialang.org/) package that implements structure-preserving numerical methods for dispersive shallow water models.
 To date, it provides provably conservative, entropy-conserving and well-balanced numerical schemes for some dispersive shallow water models:
 
+* the [Korteweg–De Vries (KdV) equation the prototypical example of an integrable PDE](https://doi.org/10.1007/s10915-025-02898-x)
```
JoshuaLampert, line 15:
```suggestion
* the [Korteweg–De Vries (KdV) equation as the prototypical example of an integrable PDE](https://doi.org/10.1007/s10915-025-02898-x),
```

JoshuaLampert (changes requested):

cwittens:
> > > > > > And I am not sure I did add OrdinaryDiffEqRosenbrock in the right way (?)
> > > > > 
> > > > > 
> > > > > Looks good.
> > > > 
> > > > 
> > > > Why are there errors in the Downgrade CI then?
> > > 
> > > 
> > > Ah, we need to bump the version to resolve the unsatisfiable requirements.
> > 
> > 
> > how / where do I do that?
> 
> So the compat bound for OrdinaryDiffEqRosenbrock.jl [is set to v1.1](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/198/files#diff-ad1f158d5c6f62a7fe387246b89370948e00b895a395dd451ab5106932fcef62R22). The Downgrade action automatically tests whether the compat bounds in our Project.toml and test/Project.toml are set correctly by setting (downgrading) the compat bound to the lowest supported version (so `OrdinaryDiffEqRosenbrock = "1.1"` becomes `OrdinaryDiffEqRosenbrock = "=1.1"`), which means it tries to run the tests with all versions as the minimal ones that are allowed by our Project.toml files. What the error tells you, is that OrdinaryDiffEqRosenbrock.jl@1.1 is not compatible with the other compat bounds because there are packages with the minimal version allowed by our Project.toml that require certain versions of other packages, which are not allowed by OrdinaryDiffEqRosenbrock.jl. In this case
> 
> * OrdinaryDiffEqRosenbrock.jl@1.1 [requires at least LinearSolve.jl@2.32](https://github.com/SciML/OrdinaryDiffEq.jl/blob/7a98e43b63691a14e2e3e80f803fa324bee83ad4/lib/OrdinaryDiffEqRosenbrock/Project.toml#L34)
> * LinearSolve.jl >= v2.23 (in particular v2.32) [requires at least BandedMatrices.jl v1.5](https://github.com/SciML/LinearSolve.jl/blob/v2.23.0/Project.toml#L64)
> * [We allow BandedMatrices.jl v1](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/8e2f9463bb613bc3fd31230610c7cb28f6602bd3/Project.toml#L29),which downgrades to `BandedMatrices=1.0` as compat bound
> 
> This means that on the one hand we need at least BandedMatrices.jl=1.0 from our compat, but on the other hand BandedMatrices>=1.5 indirectly coming from OrdinaryDiffEqRosenbrock.jl=1.1. There are two fixes for this:
> 
> 1. Either OrdinaryDiffEqRosenbrock.jl (and therefore LinearSolve.jl) allows lower versions of BandedMatrices.jl (specifically v1.0 in our case). I think this would theoretically be possible (not sure about this specific case, but I encountered similar problems, where I think it should), but the SciML people tend to just set pretty high (often unnecessarily high) compat bounds and don't care this breaks Downgrade actions (while at the same time [recommending Downgrade](https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatibility/110063) (well...and [also not recommending it](https://github.com/SciML/OrdinaryDiffEq.jl/issues/2639#issuecomment-2760801931))).
> 2. We bump our compat bound of BandedMatrices.jl to v1.5. This is of course the way to do because it's easy to do and doesn't require fighting with the whole SciML ecosystem.
> 
> But, yeah, generally Downgrade tends to be very annoying, but IMHO enhances code quality.

Thank you for the detailed explanation.
I really dont feel like fighting right now, so (I think) I bumped up BandedMatrices.jl to v1.5. and hope all the tests will pass now.

cwittens:
I now also changed PreallocationTools = "0.4.14" to "0.4.27", because I think this is what the error message told me to.
This again also fails. I am not sure what to do here.
Besides that the PR is finished from what I can tell.

JoshuaLampert:
"Want" is a bit exaggerated, but I can have a look later 😅

Project.toml
```diff
@@ -26,14 +26,14 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
...
-BandedMatrices = "1"
+BandedMatrices = "1.5"
 DiffEqBase = "6.157"
 FastBroadcast = "0.3.5"
 ForwardDiff = "0.10.36"
 Interpolations = "0.15.1"
 LinearAlgebra = "1"
 PolynomialBases = "0.4.15"
-PreallocationTools = "0.4.14"
+PreallocationTools = "0.4.27"
```
ranocha, line 36:
```suggestion
PreallocationTools = "0.4.23"
```

This is what https://github.com/SciML/OrdinaryDiffEq.jl/blame/master/lib/OrdinaryDiffEqNonlinearSolve/Project.toml requires

docs/src/overview.md
```diff
@@ -73,8 +73,8 @@ far to the left in order to avoid an interaction of the left- and right-travelin
 
 In the next step, we build a [`Semidiscretization`](@ref) that bundles all ingredients for the spatial discretization of the model. Especially, we need to
 define a [`Solver`](@ref). The simplest way to define a solver is to call the constructor by providing the mesh and a desired order of accuracy. In the
-following example, we use an accuracy order of 4. The default constructor simply creates periodic first- and second-derivative central finite difference
-summation-by-parts (SBP) operators of the provided order of accuracy. How to use other summation-by-parts operators, is described in the section on
+following example, we use an accuracy order of 4. The default constructor simply creates periodic first-, second- and third-derivative central finite
```
ranocha, line 76:
```suggestion
following example, we use an accuracy order of 4. The default constructor simply creates periodic first-, second-, and third-derivative central finite
```
Oxford comma

src/equations/kdv_1d.jl
ranocha: I do not have a strong preference. I have often used `c` so far.

Project.toml
```diff
@@ -26,14 +26,14 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 TrixiBase = "9a0f1c46-06d5-4909-a5a3-ce25d3fa3284"
 
 [compat]
-BandedMatrices = "1"
+BandedMatrices = "1.5"
 DiffEqBase = "6.157"
```
ranocha, line 30: I think we need to update this to remove the nonlinear solver dependencies.

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,179 @@
...
+    cache = (; D1, D3, c_0, c_1, DD, eta2, eta2_x, eta_x, eta_xxx,)
+    return cache
+end
+
+function rhs!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
+              ::BoundaryConditionPeriodic, source_terms, solver, cache)
+    eta, = q.x
+    deta, = dq.x
+
+    (; D1, D3, c_0, c_1, DD,) = cache
```
JoshuaLampert, line 146:
```suggestion
    (; D1, D3, c_0, c_1, DD) = cache
```

src/solver.jl
```diff
@@ -13,58 +13,78 @@ A `struct` that holds the summation-by-parts (SBP) operators that are used for t
...
+    Solver{real(D1), typeof(D1), typeof(D2), typeof(D3)}(D1, D2, D3)
 end
 
 # Also allow to pass custom SBP operators (for convenience without explicitly specifying the type)
 """
-    Solver(D1, D2)
+    Solver(D1, D2 = nothing, D3 = nothing)
 
-Create a solver, where `D1` is an `AbstractDerivativeOperator`
+Create a solver, where `D1` is an `AbstractDerivativeOperator`
```
JoshuaLampert, line 74:
```suggestion
Create a solver, where `D1` is an `AbstractDerivativeOperator`
```

src/solver.jl
```diff
@@ -13,58 +13,78 @@ A `struct` that holds the summation-by-parts (SBP) operators that are used for t
...
-Create a solver, where `D1` is an `AbstractDerivativeOperator`
+Create a solver, where `D1` is an `AbstractDerivativeOperator` 
 from [SummationByPartsOperators.jl](https://github.com/ranocha/SummationByPartsOperators.jl)
-of first `derivative_order` and `D2` is an `AbstractDerivativeOperator`
-of second `derivative_order` or an `AbstractMatrix`. It can also be `nothing`
-if no second derivative is used by the discretization.
-Both summation-by-parts operators should be associated with the same grid.
+of first `derivative_order`, `D2` is an `AbstractDerivativeOperator` 
+of second `derivative_order` or an `AbstractMatrix`, and `D3` is an `AbstractDerivativeOperator`
+of third `derivative_order` or an `AbstractMatrix`. Each can also be `nothing`
```
JoshuaLampert, line 78:
```suggestion
of third `derivative_order` or an `AbstractMatrix`. `D2` and `D3` can also be `nothing`
```

JoshuaLampert (changes requested):
Thanks again! This is close to be ready to be merged. Please find a final round of review below.

src/equations/kdv_1d.jl
JoshuaLampert: I tend to use `c`. This probably means we need to update the test values. Sorry :grimacing:

cwittens: "we" will do that now 🥲

src/DispersiveShallowWater.jl
```diff
@@ -64,9 +64,11 @@ include("util.jl")
 export examples_dir, get_examples, default_example, convergence_test
 
 export AbstractShallowWaterEquations,
+       KdVEquation1D,
        BBMEquation1D, BBMBBMEquations1D,
        SvärdKalischEquations1D, SvaerdKalischEquations1D,
        SerreGreenNaghdiEquations1D, HyperbolicSerreGreenNaghdiEquations1D
+
```
JoshuaLampert, line 71:
```suggestion
```

JoshuaLampert (changes requested):
I have three more requests:
1. Can you please build the documentation locally to see if everything (especially the formulas in the docstrings) renders fine? How to do that is explained [in the docs](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/development/#Preview-of-the-documentation).
2. Can you provide a convergence study again?
3. Please open an issue with the TODOs in your initial comment of this PR so that we can better keep track of the progress regarding the KdV equation.

cwittens:
EOC test:
For the manufactured solution EOC of 4 (up to 512)
```julia
l2
η
N    error     EOC
32   5.76e-03  -
64   3.28e-04  4.13
128  2.05e-05  4.00
256  1.29e-06  4.00
512  8.19e-08  3.97
1024 8.93e-08  -0.12

mean           3.20
linf
η
N    error     EOC
32   5.83e-03  -
64   3.44e-04  4.08
128  2.09e-05  4.04
256  1.31e-06  3.99
512  8.73e-08  3.91      
1024 7.90e-08  0.14

mean           3.23
```


For the solitary wave solution EOC of about 3:
```julia
l2
η
N    error     EOC
64   6.70e-01  -
128  6.72e-02  3.32
256  1.29e-02  2.38
512  1.11e-03  3.54

mean           3.08
linf
η
N    error     EOC
64   3.96e-01  -
128  4.61e-02  3.10
256  1.04e-02  2.14
512  8.58e-04  3.60

mean           2.95
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,183 @@
...
+    return SVector(eta)
+end
+
+"""
+    source_terms_manufactured(q, x, t, equations::KdVEquation1D, mesh)
+
+A smooth manufactured solution in combination with [`initial_condition_manufactured`](@ref).
+
+How it was calculated, is described in:
+https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/198/files#r2090805751
```
JoshuaLampert, line 89:
```suggestion
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/198#discussion_r2090805751
```
The other link didn't lead me directly to the relevant comment.

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,183 @@
...
+where ``D_1`` is a first-derivative operator, ``D_3`` a third-derivative operator, and ``D`` the still-water depth.
+
+It conserves
+- the total water mass (integral of ``\eta``) as a linear invariant
+and if upwind operators (``D_3 = D_{1,+} D_1 D_{1,-}``) or wide-stencil operators (``D_3 = D_1^3``) are used for the third derivative, it also conserves
+- the energy (integral of ``1/2\eta^2``)
+
+for periodic boundary conditions.
+
+- Diederik Korteweg and Gustav de Vries (1895).
```
JoshuaLampert, line 38:
```suggestion
- Diederik Korteweg and Gustav de Vries (1895)
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,183 @@
...
+"""
+    initial_condition_convergence_test(x, t, equations::KdVEquation1D, mesh)
+
+A traveling-wave solution used for convergence tests in a periodic domain, here for dimensional variables.
+"""
+function initial_condition_convergence_test(x, t, equations::KdVEquation1D, mesh)
+    g = gravity(equations)
+    D = equations.D
+    c = 1.5 * sqrt(g * D)
+    A = 2 * D * (c - sqrt(g * D)) / sqrt(g * D)
```
JoshuaLampert, line 65:
```suggestion
    c0 = sqrt(g * D)
    c = 1.5 * c0
    A = 2 * D * (c - c0) / c0
```

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,183 @@
...
+    eta2_x = DiffCache(zero(template), N)
+
+    eta_x = DiffCache(zero(template), N)
+    eta_xxx = DiffCache(zero(template), N)
+
+    if solver.D1 isa PeriodicUpwindOperators
+        D1 = solver.D1.central
+
+        # calculate the third derivative operator using upwind operators
+        D3 = sparse(solver.D1.plus) * sparse(solver.D1.minus) * sparse(solver.D1.central)
```
JoshuaLampert, line 135:
```suggestion
        D3 = sparse(solver.D1.plus) * sparse(solver.D1.central) * sparse(solver.D1.minus)
```
Don't we need the central one in the middle to prove semidiscrete energy conservation? I fear with this change, some test values may change a bit again. Sorry for missing that before :grimacing:

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,183 @@
...
+    else
+        D1 = solver.D1
+        D3 = solver.D3
+    end
+
+    cache = (; D1, D3, c_0, c_1, DD, eta2, eta2_x, eta_x, eta_xxx,)
```
JoshuaLampert, line 141:
```suggestion
    cache = (; D1, D3, c_0, c_1, DD, eta2, eta2_x, eta_x, eta_xxx)
```

JoshuaLampert (changes requested):
Final final minor comments and one question.

src/equations/kdv_1d.jl
cwittens: ohh yes, this is a blunder on my side

test/test_kdv_1d.jl
```diff
@@ -0,0 +1,61 @@
...
+                        tspan=(0.0, 5.0),
+                        l2=[0.0011470884314204559],
+                        linf=[0.0008506440340751986],
+                        cons_error=[3.1086244689504383e-15],
+                        change_waterheight=3.1086244689504383e-15)
+
+    @test_allocations(semi, sol, allocs=5_000)
+end
+
+@testitem "kdv_1d_basic using ForwardDiff" setup=[Setup, KdVEquation1D] begin
```
JoshuaLampert, line 16:
```suggestion
@testitem "kdv_1d_basic_implicit" setup=[Setup, KdVEquation1D] begin
```

JoshuaLampert (dismissed):
Thanks a lot!

src/equations/kdv_1d.jl
```diff
@@ -0,0 +1,184 @@
...
+    eta2_x = DiffCache(zero(template), N)
+
+    eta_x = DiffCache(zero(template), N)
+    eta_xxx = DiffCache(zero(template), N)
+
+    if solver.D1 isa PeriodicUpwindOperators
+        D1 = solver.D1.central
+
+        # calculate the third derivative operator using upwind operators
+        D3 = sparse(solver.D1.plus) * sparse(solver.D1.central) * sparse(solver.D1.minus)
```
ranocha, line 136: Why do you store the sparse matrix instead of applying the operators consecutively?

examples/kdv_1d/kdv_1d_narrow_stencil.jl
```diff
@@ -0,0 +1,41 @@
...
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 4
+
+# using a narrow stencil third derivative operator is significantly faster than using
+# the upwind operator for the third derivative for a given N. However there are considerably more oscillations.
+# This can be improved by increasing N, which in fact leads to it being slower again.
+# TODO: This text maybe in the docs somewhere?
+accuracy_order = 4
+solver = Solver(mesh, accuracy_order)
```
ranocha, line 24: Is this because you construct a sparse matrix for `D3`?

examples/kdv_1d/kdv_1d_manufactured.jl
```diff
@@ -0,0 +1,47 @@
...
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 4
+accuracy_order = 4
+D1_upwind = upwind_operators(periodic_derivative_operator;
+                             derivative_order = 1, accuracy_order = 4,
+                             xmin = xmin(mesh), xmax = xmax(mesh),
+                             N = nnodes(mesh))
+
+solver = Solver(D1_upwind, nothing)
```
ranocha, line 27:
```suggestion
# create solver with periodic SBP operators of accuracy order 4;
# we can set the accuracy order of the upwind operators to 3 since
# we only use central versions/combinations of the upwind operators
D1_upwind = upwind_operators(periodic_derivative_operator;
                             derivative_order = 1, accuracy_order = 3,
                             xmin = xmin(mesh), xmax = xmax(mesh),
                             N = nnodes(mesh))
solver = Solver(D1_upwind)
```

examples/kdv_1d/kdv_1d_implicit.jl
```diff
@@ -0,0 +1,42 @@
...
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic upwind SBP operators of accuracy order 4
+accuracy_order = 4
+D1_upwind = upwind_operators(periodic_derivative_operator;
+                             derivative_order = 1, accuracy_order = 4,
+                             xmin = xmin(mesh), xmax = xmax(mesh),
+                             N = nnodes(mesh))
+
+solver = Solver(D1_upwind)
```
ranocha, line 25:
```suggestion
# create solver with periodic SBP operators of accuracy order 4;
# we can set the accuracy order of the upwind operators to 3 since
# we only use central versions/combinations of the upwind operators
D1_upwind = upwind_operators(periodic_derivative_operator;
                             derivative_order = 1, accuracy_order = 3,
                             xmin = xmin(mesh), xmax = xmax(mesh),
                             N = nnodes(mesh))
solver = Solver(D1_upwind)
```

examples/kdv_1d/kdv_1d_basic.jl
```diff
@@ -0,0 +1,42 @@
...
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic upwind SBP operators of accuracy order 4
+accuracy_order = 4
+D1_upwind = upwind_operators(periodic_derivative_operator;
+                             derivative_order = 1, accuracy_order = 4,
+                             xmin = xmin(mesh), xmax = xmax(mesh),
+                             N = nnodes(mesh))
+
+solver = Solver(D1_upwind)
```
ranocha, line 25:
```suggestion
# create solver with periodic SBP operators of accuracy order 4;
# we can set the accuracy order of the upwind operators to 3 since
# we only use central versions/combinations of the upwind operators
D1_upwind = upwind_operators(periodic_derivative_operator;
                             derivative_order = 1, accuracy_order = 3,
                             xmin = xmin(mesh), xmax = xmax(mesh),
                             N = nnodes(mesh))
solver = Solver(D1_upwind)
```

ranocha (commented):
Thanks!

src/equations/kdv_1d.jl
cwittens: I did that first. I thought doing one expanse calculation at the start once is faster in the long run.
Also it is done like this in the Documentation (for 2nd derivate) https://numericalmathematics.github.io./DispersiveShallowWater.jl/stable/overview/#customize_solver

ranocha: That's because we need to solve linear systems with the second-derivative operators. Did you compare the runtime with just applying the upwind operators consecutively?

examples/kdv_1d/kdv_1d_narrow_stencil.jl
cwittens: I also benchmarked now with fixed time steps:
```julia
# for abstol = 1e-7, reltol = 1e-7:
# 19.288 ms (5359 allocations: 718.70 KiB) narrow stencil
# 42.618 ms (6193 allocations: 751.51 KiB) upwind

# for fixed time steps (dt = 0.01):
# 17.194 ms (1838 allocations: 650.54 KiB) narrow stencil
# 32.263 ms (1838 allocations: 649.72 KiB) upwind
```
but the solution is qualitatively worse for the narrow stencil.
image
image

src/equations/kdv_1d.jl
cwittens: Should I check runtime including creating the cache? So basically the whole Elixir? 
Or just of the solve?

cwittens:
MacOs is failing me again 😭😭
I will try and fix that after having benchmarked 

> Why do you store the sparse matrix instead of applying the operators consecutively?

just in case I need to change error values again.

src/equations/kdv_1d.jl
cwittens: A quick and dirty implementation is already considerably fast when applying the operators consecutively. 
(This low key confuses me, I thought sparse matrices where only somewhat slower then SBP operators)

I will do a commit changing it tomorrow.

ranocha: Well, I would have expected roughly an order of magnitude difference based on benchmarks like https://ranocha.de/SummationByPartsOperators.jl/stable/benchmarks/#Periodic-domains

src/solver.jl
```diff
@@ -13,58 +13,78 @@ A `struct` that holds the summation-by-parts (SBP) operators that are used for t
...
+Create a solver, where the three summation-by-parts (SBP) first-, second-, and third-derivative operators
+are of accuracy order `accuracy_order` and associated to the `mesh`.
 """
 function Solver(mesh, accuracy_order)
     if isodd(accuracy_order)
         @warn "DispersiveShallowWater.jl is expecting a central difference operator. This is not given for an odd accuracy (got $accuracy_order) order.\n This can lead to a significant reduction in the order of convergence of the solution."
     end
     D1 = periodic_derivative_operator(1, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
     D2 = periodic_derivative_operator(2, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
+    D3 = periodic_derivative_operator(3, accuracy_order, mesh.xmin, mesh.xmax, mesh.N)
```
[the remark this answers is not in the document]
JoshuaLampert: That's good to know. Thanks for pointing that out. I just did a convergence test with kdv_1d_narrow_stencil.jl setting `abstol = 1e-12, reltol = 1e-12`:

current version:
```julia
julia> convergence_test("examples/kdv_1d/kdv_1d_narrow_stencil.jl", 5; N = 64)
[...]
l2
η                        
N    error     EOC       
64   1.42e+00  -         
128  1.38e-01  3.36      
256  2.74e-01  -0.99     
512  8.43e-02  1.70      
1024 2.15e-02  1.97      

mean           1.51      
linf
η                        
N    error     EOC       
64   7.02e-01  -         
128  1.04e-01  2.76      
256  2.08e-01  -1.00     
512  6.53e-02  1.67      
1024 1.66e-02  1.98      

mean           1.35
```

using `accuracy_order + 2` for the third-derivative operator:

```julia
julia> convergence_test("examples/kdv_1d/kdv_1d_narrow_stencil.jl", 5; N = 64)
[...]
l2
η                        
N    error     EOC       
64   1.47e+00  -         
128  5.16e-01  1.51      
256  4.68e-02  3.46      
512  2.96e-03  3.98      
1024 1.84e-04  4.00      

mean           3.24      
linf
η                        
N    error     EOC       
64   7.38e-01  -         
128  3.17e-01  1.22      
256  3.31e-02  3.26      
512  2.08e-03  3.99      
1024 1.29e-04  4.01      

mean           3.12      
```

So I think we should use `accuracy_order + 2`. Do you understand the bad result with `N = 256` in the first case?

cwittens:
it is possible to do the ``rhs!`` with just two instead of four cache vectors, e.g.
```julia
    @trixi_timeit timer() "hyperbolic" begin
        @.. tmp_1 = eta^2
        # eta2_x = D1 * eta2
        mul!(tmp_2, D1, tmp_1)

        # eta_x = D1 * eta
        mul!(tmp_1, D1, eta)

        @.. deta = -1.0 * (c_0 * tmp_1 +
                           c_1 * (eta * tmp_1 +
                                  tmp_2))
    end
```
and so on ...

but it comes at the coast of readability. What is preferred here?

src/solver.jl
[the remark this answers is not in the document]
ranocha: Bad luck in the non-symptotic regime?

ranocha:
That's what I have done often enough so far 🙈 It's fine with me if the complexity is local enough.

JoshuaLampert (changes requested):
LGTM! I have only a few minor suggestions.

cwittens:
This CI fails for MacOs where for the BBM manufactured solution (allocs 5024 < 5000 error) and not for the KdV manufactured solution. 
I think the source of the error was the change to ``D3 = periodic_derivative_operator(3, accuracy_order + 2, mesh.xmin, mesh.xmax, mesh.N)``. Maybe D3 got slightly bigger and with this it came just over the allocations allowed. IDK.

JoshuaLampert:
> This CI fails for MacOs where for the BBM manufactured solution (allocs 5024 < 5000 error) and not for the KdV manufactured solution. I think the source of the error was the change to `D3 = periodic_derivative_operator(3, accuracy_order + 2, mesh.xmin, mesh.xmax, mesh.N)`. Maybe D3 got slightly bigger and with this it came just over the allocations allowed. IDK.

Hm, but `D3` doesn't appear in the `rhs!` of the BBM equation and `@test_allocations` [only looks for allocations in `rhs!`](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/386221cbfa9a84b9611c58d18fb19555d3ce6b1a/test/test_util.jl#L175).

JoshuaLampert (dismissed):

examples/kdv_1d/kdv_1d_basic.jl
```diff
@@ -0,0 +1,42 @@
...
+N = 512
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 3;
+# we can set the accuracy order of the upwind operators to 3 since
+# we only use central versions/combinations of the upwind operators
+D1_upwind = upwind_operators(periodic_derivative_operator;
+                             derivative_order = 1, accuracy_order = 3,
+                             xmin = xmin(mesh), xmax = xmax(mesh),
+                             N = nnodes(mesh))
```
ranocha, line 24:
```suggestion
# create solver with periodic SBP operators of accuracy order 4;
# we can set the accuracy order of the upwind operators to 3 since
# we only use central versions/combinations of the upwind operators
D1_upwind = upwind_operators(periodic_derivative_operator;
                             derivative_order = 1, accuracy_order = 3,
                             xmin = xmin(mesh), xmax = xmax(mesh),
                             N = nnodes(mesh))
```

examples/kdv_1d/kdv_1d_implicit.jl
```diff
@@ -0,0 +1,42 @@
...
+boundary_conditions = boundary_condition_periodic
+
+# create homogeneous mesh
+coordinates_min = -50.0
+coordinates_max = 50.0
+N = 512
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 3;
+# we can set the accuracy order of the upwind operators to 3 since
```
ranocha, line 19:
```suggestion
# create solver with periodic SBP operators of accuracy order 4;
# we can set the accuracy order of the upwind operators to 3 since
```

ranocha (commented):
Thanks! Just a minor round of suggestions - we should mention that the resulting discretizations are fourth-order accurate, not third-order, shall we?

cwittens:
> Thanks! Just a minor round of suggestions - we should mention that the resulting discretizations are fourth-order accurate, not third-order, shall we?

I wrote it down explicitly, because just changing it to "3" would make me think it is just a typo (as I did before).

examples/kdv_1d/kdv_1d_basic.jl
```diff
@@ -0,0 +1,43 @@
...
+# create homogeneous mesh
+coordinates_min = -50.0
+coordinates_max = 50.0
+N = 512
+mesh = Mesh1D(coordinates_min, coordinates_max, N)
+
+# create solver with periodic SBP operators of accuracy order 3,
+# which results in a 4th order accurate semi discretizations.
+# we can set the accuracy order of the upwind operators to 3 since
+# we only use central versions/combinations of the upwind operators
```
JoshuaLampert, line 21:
```suggestion
# we can set the accuracy order of the upwind operators to 3 since
# we only use central versions/combinations of the upwind operators
```

ranocha (approved):
Thanks!

PR #201, opened 2025-05-16
Document `plot_initial = true`

### LINKED ISSUE #199: `plot_initial` may be confusing
`plot(plot_semi => sol, plot_initial = true)` plots the initial condition passed to the semidiscretization at the *final* time. This can be confusing. Maybe we should rename this argument to `plot_analytical` or `plot_exact` or so?

JoshuaLampert:
Closes #199.


ranocha (approved):
Thanks!

PR #205, opened 2025-05-21
reflecting wall boundary conditions for Serre-Green-Naghdi

ranocha:
I have implemented reflecting boundary conditions for the Serre-Green-Naghdi equations. This version employs variable-coefficient second-derivative operators from Mattsson (2012) and conserves the modified energy.

### Energy conservation

```julia
julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_reflecting.jl", abstol = 1.0e-14, reltol = 1.0e-14); plot(semi => sol, plot_initial = true)
[...]

julia> integrals(analysis_callback).entropy_modified |> extrema |> x -> last(x) - first(x)
9.663381206337363e-13
```

### Convergence results

```julia
julia> convergence_test("examples/serre_green_naghdi_1d/serre_green_naghdi_manufactured_reflecting.jl", [64, 128, 256, 512, 1024], accuracy_order = 2, coordinates_min = 0, coordinates_max = 1, abstol = 1.0e-14, reltol = 1.0e-14, interval = 0);
[...]
l2
64   4.80e-02  -         64   2.74e-04  -         64   0.00e+00  -         
128  1.69e-02  1.51      128  6.77e-05  2.02      128  0.00e+00  NaN       
256  5.95e-03  1.50      256  1.69e-05  2.01      256  0.00e+00  NaN       
512  2.10e-03  1.50      512  4.20e-06  2.00      512  0.00e+00  NaN       
1024 7.42e-04  1.50      1024 1.05e-06  2.00      1024 0.00e+00  NaN       

mean           1.50      mean           2.01      mean           NaN       
linf
64   3.75e-01  -         64   4.75e-04  -         64   0.00e+00  -         
128  1.92e-01  0.97      128  1.16e-04  2.03      128  0.00e+00  NaN       
256  9.71e-02  0.98      256  2.87e-05  2.02      256  0.00e+00  NaN       
512  4.88e-02  0.99      512  7.15e-06  2.01      512  0.00e+00  NaN       
1024 2.45e-02  1.00      1024 1.78e-06  2.00      1024 0.00e+00  NaN       

mean           0.98      mean           2.01      mean           NaN       
```

The convergence rates are in accordance with the results of Mitsotakis et al. (reference in the docstring of the `initial_condition_manufactured_reflecting`).

Please note that a soliton is not reflected perfectly - we get a dispersive tail instead. This is also described in the reference mentioned above.

CC @cwittens @vincmarks You may use this for your project work.


ranocha:
> 2\. I see you opted for a different approach to implement the reflecting boundary conditions than we have for the Svärd-Kalisch equations (basically storing a $N \times (N - 2)$ matrix vs. a $N \times N$ matrix setting the first and last rows and columns of `A` to zero). Is this because of the variable coefficient operator or could we also use the other approach with the `Pd` matrix for the SGN equations (and vice versa)? If the latter, what are the differences and (dis-)advantages of the two approaches?

We need to assemble the banded matrix for every `rhs!` call for SGN. This is different from the other equations, where we can either use the same matrix (BBM-BBM) or at least a significant portion of it (SK). Thus, I wanted to keep a cache for the banded matrix assembly. Then, it was easier from my point of view to just reuse this banded matrix for the linear solves instead of reducing its size slightly.

ranocha:
> 1. If I understand correctly, in the reference paper they observe the dispersive tail for the reflecting soliton only for amplitudes larger than `A = 0.65`. Our soliton has an amplitude of `A = h2 - h1 = 0.2`, but we still observe the dispersive tail (they also use `b = -1`, while we have `b = 0`). Is my understanding correct and if yes, what is your opinion on that?

The choice of the bottom is just a vertical adjustment and should not matter as long as the water height at rest is the same(unity in both setups). 
I could not find an explicit statement that they did not observe the dispersive tails for amplitudes smaller than `A = 0.65` - only some jet-like structures during the interaction require these amplitudes.
I also checked smaller amplitudes, such as `A = 0.075` in got results similar to their Figure 6: I could not see a dispersive tail in the water height (shown also in their plots) but only in the velocity (not shown/discussed in the paper).

SGN

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -911,6 +1020,156 @@ function rhs_sgn_upwind!(dq, q, t, equations, source_terms, solver, cache,
...
+    A[:, begin] .= 0
+    A[begin, begin] = 1
+    A[end, :] .= 0
+    A[:, end] .= 0
+    A[end, end] = 1
+    return Symmetric(A)
+end
+
+function rhs!(dq, q, t, mesh,
+              equations::SerreGreenNaghdiEquations1D,
```
JoshuaLampert, line 1087: > For a variable bathymetry, we will need to discretize more terms, and I would like to check first by hand how they fit together for energy conservation. Thus, I have restricted this PR to the flat case and would like to implement variable bathymetry in another PR (also to keep the number of changes per PR small enough for review).

Sounds good to me. In this case, would it make sense to also restrict the `rhs!` to only flat bottom?
```suggestion
              equations::SerreGreenNaghdiEquations1D{BathymetryFlat},
```

JoshuaLampert (commented):
> We need to assemble the banded matrix for every rhs! call for SGN. This is different from the other equations, where we can either use the same matrix (BBM-BBM) or at least a significant portion of it (SK). Thus, I wanted to keep a cache for the banded matrix assembly. Then, it was easier from my point of view to just reuse this banded matrix for the linear solves instead of reducing its size slightly.

Ok, thanks for the explanation.

> I could not find an explicit statement that they did not observe the dispersive tails for amplitudes smaller than A = 0.65 - only some jet-like structures during the interaction require these amplitudes.
I also checked smaller amplitudes, such as A = 0.075, and obtained results similar to those in their Figure 6. I could not observe a dispersive tail in the water height (as shown in their plots) but only in the velocity (not shown or discussed in the paper).

Ah, I thought the jet-like structures and the dispersive tail were the same. But that was a misunderstanding from my side. Thanks for the clarification and the additional plot for `A = 0.075`, which looks good.

JoshuaLampert (approved):
Thanks a lot!

PR #206, opened 2025-05-21
Rename source term variables consistently

JoshuaLampert:
I renamed all source terms as `s1`, `s2`, ... for consistency and the `d` at the beginning of the variable names before indicated it's the time derivative `q1` , `h`, ..., but it's only one part of it (for equations involving elliptic systems it's also the source term before applying the elliptic operator). So I think it was a bit confusing. See also https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/198.
On the way I also replaced some more expressions to avoid redundant computations and make the code a bit shorter.


ranocha (approved):
Thanks

Issue #202, opened 2025-05-16
Roadmap KdV eq.


cwittens:
The following I want to do next regarding the KdV eq.

- [x] Add KdV to Benchmark results, see #208 
- [x] add nondim2prim + prim2nondim functions to be able to compare results from $$u_t + u u_x + u_{xxx}=0$$ PDE to the current implementation $$\eta_t+\sqrt{g D} \eta_x+3 / 2 \sqrt{g / D} \eta \eta_x+1 / 6 \sqrt{g D} D^2 \eta_{x x x} = 0$$, see #217
- [x] add the linear dispersion relation $\omega(k) = \sqrt{g h_0}k(1 + \frac 3 2 \frac{\eta_0}{h_0}) - \frac 1 6 h_0^2k^3$ to https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/a5e6adb4319151d9b20139005bc3f95334f8e7b7/src/dispersion_relation.jl, see #230.
- [x] add more conversion_functions, right now only waterheight is working, see #231
- [x] possibly add more examples, see #231
- [x] more unit tests, see #231
- [x] more docs
- [ ] split up the ode - make it possible to use IMEX in DSW.jl, see also #165

JoshuaLampert:
Regarding the conversion functions: I think this basically boils down to implementing `energy_total`, which should return $$\frac 1 2\eta^2$$ and `entropy`, which computes the same (just forwarding to `energy_total` similarly to `AbstractShallowWaterEquations`). This is what we would like to see conserved by the semidiscretization, i.e. checking this by adding `entropy` to the `AnalysisCallback`s of some examples would be good and maybe adding another example using relaxation for fully-discrete entropy-conservation. Finally, we should extend the unit tests for the KdV equation similarly to the BBM equation
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/a5e6adb4319151d9b20139005bc3f95334f8e7b7/test/test_unit.jl#L131-L177
Functions like `prim2prim`, `prim2cons`, and `prim2phys` should already work out of the box. The others (`energy_total_modified`, `entropy_modified`, and `hamiltonian`) should not be needed for the KdV equation.

PR #207, opened 2025-05-23
Added IMEX for KdV eq.

cwittens:
I wished I could have done something like
```julia
function rhs!(...)
   rhs_split_1!(...)
   rhs_split_2!(...)
end
```
but this does not seem practically.

see comments in:
```julia
function rhs!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
              ::BoundaryConditionPeriodic, source_terms, solver, cache)
    # define locally bc to not use a global variable
    bc = boundary_condition_periodic      
    (; tmp1) = cache
    rhs_1!(dq, q, t, mesh, equations, initial_condition,
           bc, source_terms, solver, cache)

    deta, = dq.x
    tmp1 .= deta

    # rhs_split_2! needs to be defined to set "deta = ..."
    # and not just "deta += ..." in order to work with SplitODEProblem 
    # This means one either has to store the result in a temporary variable
    # or have a function rhs! does not call rhs_split_1! and rhs_split_2!. 
    # and currently it does not work with ForwardDiff
    rhs_split_2!(dq, q, t, mesh, equations, initial_condition,
           bc, source_terms, solver, cache)

    deta .+= tmp1
    return nothing
end
```

Here I am not 100% if this works, but I think it does (and tests will tell I hope).
```julia
function jacobian(semi::Semidiscretization;
                  t = 0.0,
                  q0 = compute_coefficients(semi.initial_condition, t, semi))
-    J = ForwardDiff.jacobian(similar(q0), q0) do dq, q
-         DispersiveShallowWater.rhs!(dq, q, semi, t)
+    @unpack tmp_partitioned = semi.cache
+    J = ForwardDiff.jacobian(tmp_partitioned, q0) do dq, q
+        DispersiveShallowWater.rhs!(dq, q, semi, t)
    end
    return J
```


examples/kdv_1d/kdv_1d_IMEX.jl
```diff
@@ -0,0 +1,46 @@
...
+summary_callback = SummaryCallback()
+analysis_callback = AnalysisCallback(semi; interval = 100,
+                                     extra_analysis_errors = (:conservation_error,),
+                                     extra_analysis_integrals = (waterheight_total,
+                                                                 waterheight))
+callbacks = CallbackSet(analysis_callback, summary_callback)
+saveat = range(tspan..., length = 100)
+
+# alg = Rodas5() # not working because of https://github.com/SciML/OrdinaryDiffEq.jl/issues/2719
+# alg = KenCarp4() # would need to add OrdinaryDiffEqSDIRK - can to that, but I dont want to bloat DSW.jl with package s
```
JoshuaLampert, line 43: I would prefer to use a true IMEX scheme like `KenCarp4` here. Something like `Tsit5` will use explicit time stepping also for the stiff part, right? I think it's fine to just add OrdinaryDiffEqSDIRK.jl to the test suite. It doesn't need to be a dependency of DSW.jl.

src/semidiscretization.jl
```diff
@@ -214,16 +236,21 @@ function check_bathymetry(equations::AbstractShallowWaterEquations, q0)
...
 
 """
-    semidiscretize(semi::Semidiscretization, tspan)
+    semidiscretize(semi::Semidiscretization, tspan; no_splitform = true)
 
 Wrap the semidiscretization `semi` as an ODE problem in the time interval `tspan`
 that can be passed to `solve` from the [SciML ecosystem](https://diffeq.sciml.ai/latest/).
 """
-function semidiscretize(semi::Semidiscretization, tspan)
+function semidiscretize(semi::Semidiscretization, tspan; no_splitform = true)
```
JoshuaLampert, line 244: I'm wondering what's the best interface for this. I [initially thought](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/issues/165#issuecomment-2518708858) it would be good to add a trait `has_stiff_terms(equations)`, which can be used to determine whether we create a usual `ODEFunction` or a `SplitFunction`. This way, we wouldn't need a kwarg `no_splitform` or similar, but whether or not using the `SplitFunction` formulation will be determined from the equations. However, this would mean that we also use a split function in the case of a non-IMEX scheme, which could have some overhead (I think that's a similar approach as Trixi.jl does with periodic equations, right @ranocha?). Maybe you could first add one or two KdV examples to the benchmark suite as in the list of #202, such that we can try how much overhead we would have from the `SplitFunction` formulation for the explicit solvers?

src/equations/kdv_1d.jl
```diff
@@ -181,3 +181,72 @@ function rhs!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
 
     return nothing
 end
+
+function rhs_split_1!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
```
JoshuaLampert, line 185:
```suggestion
function rhs_split_stiff!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
```
I think that's more descriptive.

src/equations/kdv_1d.jl
```diff
@@ -181,3 +181,72 @@ function rhs!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
...
+
+        # deta = 1 / 6 sqrt(g * D) D^2 eta_xxx 
+
+        @.. deta = -1 / 6 * c_0 * DD * tmp_1
+    end
+
+    return nothing
+end
+
+function rhs_split_2!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
```
JoshuaLampert, line 216:
```suggestion
function rhs_split_nonstiff!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
```
See above.

JoshuaLampert (commented):
Just an initial review with a general question about the interface.

src/semidiscretization.jl
JoshuaLampert: The advantage of the `has_stiff_terms` (or maybe better `have_stiff_terms` because `equations` are plural) would be:
- the user would not need to set `no_splitform = false` in `semidiscretize` in order to use an IMEX scheme
- we wouldn't need the `rhs!` definition anymore because we always define a `SplitFunction` for equations, which `have_stiff_terms` reducing the maintenance overhead and avoiding duplication.

The disadvantage I see is
- maybe a small overhead.

Any more aspects to consider?

ranocha: Hyperbolic approximations would be tricky since they can have quite stiff terms depending on the hyperbolic approximation parameter (`lambda` for the hyperbolic SGN equations). Still, we may also want to use them with explicit time integration methods as efficiently as possible.
Currently, "as efficiently as possible" is significantly restricted by https://github.com/SciML/RecursiveArrayTools.jl/issues/400, so I am unsure of its importance. We should benchmark it.
To sum up, we could add a trait `have_stiff_terms` to the `equations` and introduce a keyword argument like `split_ode = have_stiff_terms(equations)`. We should ensure that this results in a type-stable `semidiscretize`, i.e., the keyword argument can take `True()` and `False()` values (with different types).

ranocha:
Could you please try this a bit more - and maybe ask on Julia Slack/Discourse? We can also test using `sparse` (or `Matrix` for Fourier operators) versions if this helps to make it work first.

src/semidiscretization.jl
JoshuaLampert: Sounds good to me. So if I understand correctly, we want to support both the classical `ODEFunction` version and the `SplitODEFunction` version, i.e. we also need to keep both `rhs!` and `rhs_split_stiff!` and `rhs_split_nonstiff!`. Is there an elegant way to reduce code duplication and make maintenance easier because as Collin wrote in the initial comment of the PR we cannot simply call both versions after each other in `rhs!`.

cwittens:
> Could you please try this a bit more - and maybe ask on Julia Slack/Discourse? We can also test using `sparse` (or `Matrix` for Fourier operators) versions if this helps to make it work first.

Yes, will do.
Probably also needed because currently IMEX is not faster (see some benchmarking results in ``examples/kdv_1d/kdv_1d_manufactured.jl`` as a raw string at the end of the file.

src/semidiscretization.jl
```diff
@@ -214,18 +234,53 @@ function check_bathymetry(equations::AbstractShallowWaterEquations, q0)
...
+end
+
+function _check_split_rhs_implementation(semi)
+    @unpack mesh, equations, initial_condition, boundary_conditions, solver, source_terms, cache = semi
+
+    equation_name = get_name(equations)
+    args = (nothing, nothing, nothing, mesh, equations, initial_condition,
+            boundary_conditions, source_terms, solver, cache)
+
+    # # Check if methods are applicable
```
JoshuaLampert, line 272:
```suggestion
    # Check if methods are applicable
```

test/test_kdv_1d.jl
```diff
@@ -59,3 +59,14 @@ end
...
+
+@testitem "kdv_1d_IMEX" setup=[Setup, KdVEquation1D] begin
+    @test_trixi_include(joinpath(EXAMPLES_DIR, "kdv_1d_IMEX.jl"),
+                        tspan=(0.0, 5.0),
+                        l2=[0.004952174509850488],
+                        linf=[0.003962890861977875],
+                        cons_error=[2.220446049250313e-15],
+                        change_waterheight=-2.220446049250313e-15)
+
+    @test_allocations_splitform(semi, sol, allocs=5_000)
```
JoshuaLampert, line 71:
```suggestion
    @test_allocations_split_ode(semi, sol, allocs=5_000)
```

test/test_util.jl
```diff
@@ -176,6 +176,24 @@ macro test_allocations(semi, sol, allocs)
     end
 end
 
+"""
+    @test_allocations_splitform(semi, sol, allocs)
```
JoshuaLampert, line 180:
```suggestion
    @test_allocations_split_ode(semi, sol, allocs)
```

test/test_util.jl
```diff
@@ -176,6 +176,24 @@ macro test_allocations(semi, sol, allocs)
...
+
+Test that the memory allocations of `DispersiveShallowWater.rhs_split_stiff!` 
+and `DispersiveShallowWater.rhs_split_nonstiff!` are below `allocs`
+(e.g., from type instabilities).
+"""
+macro test_allocations_splitform(semi, sol, allocs)
```
JoshuaLampert, line 186:
```suggestion
macro test_allocations_split_ode(semi, sol, allocs)
```

src/equations/kdv_1d.jl
```diff
@@ -53,6 +53,9 @@ function KdVEquation1D(; gravity, D = 1.0, eta0 = 0.0)
     KdVEquation1D(gravity, D, eta0)
 end
 
+# KdV equations have stiff third-order derivative terms that benefit IMEX methods
```
JoshuaLampert, line 56:
```suggestion
# KdV equations have stiff third-order derivative terms that benefit from IMEX methods
```

JoshuaLampert (changes requested):
Thanks! Implementation-wise this looks good to me. I also like the way you combine the stiff and non-stiff parts in `rhs!`. Regarding performance of the IMEX method, this doesn't look good, but I don't know what's the issue. Is this related to the issue with RecursiveArrayTools.jl or do you have any idea, @ranocha?

src/equations/kdv_1d.jl
```diff
@@ -181,3 +185,135 @@ function rhs!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
 
     return nothing
 end
+
+function rhs_split_stiff!(dq, q, t, mesh, equations::KdVEquation1D, initial_condition,
```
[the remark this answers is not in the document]
JoshuaLampert: That's basically a reminder to remove the commented code once everything is settled.

ranocha:
> Thanks! Implementation-wise this looks good to me. I also like the way you combine the stiff and non-stiff parts in `rhs!`. Regarding performance of the IMEX method, this doesn't look good, but I don't know what's the issue. Is this related to the issue with RecursiveArrayTools.jl or do you have any idea, @ranocha?

The results mentioned in one of the files suggest that there could be several reasons (since the number of time steps varies, too). Thus, I propose the following steps to investigate this issue further:

- Switch to a physically relevant setup
- Compare the runtime for fixed time steps without any adaptivity
- Check a few other solvers for split ODE problems to see whether the issue persists or is special to `KenCarp4`

cwittens:
> * Compare the runtime for fixed time steps without any adaptivity

Results for fixed time steps are in the manufactured solution example below and there IMEX is better then No IMEX.

cwittens:
> * Switch to a physically relevant setup
done in https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/207#discussion_r2106233549 and it looks like IMEX is better then No IMEX, but purely explicit solvers are still better. Besides in the case of the manufactured solution, where `Tsit5()` will cause a `maxiter` exceeded error. 

> * Check a few other solvers for split ODE problems to see whether the issue persists or is special to `KenCarp4`

I tried all solvers from https://docs.sciml.ai/DiffEqDocs/stable/solvers/split_ode_solve/ besides the ``KenCarp#``, but I could really compare them. Here is a list of the problem:

### First, all Solvers:
 - `SplitEuler`: 1st order fully explicit method. Used for testing accuracy
    of splits.
  - `IMEXEuler` : 1st order explicit Euler mixed with implicit Euler. Fixed time
    step only.
  - `CNAB2`: Crank-Nicolson Adams Bashforth Order 2. Fixed time step only.
  - `CNLF`: Crank-Nicolson Leapfrog of Order 2. Fixed time step only.
  - `SBDF2` : 2nd order IMEX BDF method. Fixed time step only.
  - `SBDF3` : 3rd order IMEX BDF method. Fixed time step only. In development.
  - `SBDF4` : 4th order IMEX BDF method. Fixed time step only. In development.
  - `KenCarp3`: An A-L stable stiffly-accurate 3rd order ESDIRK method.
  - `KenCarp4`: An A-L stable stiffly-accurate 4th order ESDIRK method.
  - `KenCarp47` - An A-L stable stiffly-accurate 4th order seven-stage ESDIRK method with splitting
  - `KenCarp5`: An A-L stable stiffly-accurate 5th order ESDIRK method.
  - `KenCarp58` - An A-L stable stiffly-accurate 5th order eight-stage ESDIRK method with splitting
  - `LawsonEuler` - First order exponential Euler scheme. Fixed timestepping only.
  - `NorsettEuler` - First order exponential-RK scheme. Fixed timestepping only. Alias: `ETD1`.
  - `ETD2` - Second order Exponential Time Differencing method (in development). Fixed timestepping only. Doesn't support Krylov approximation.
  - `ETDRK2` - 2nd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK3` - 3rd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK4` - 4th order exponential-RK scheme. Fixed timestepping only.
  - `HochOst4` - 4th order exponential-RK scheme with stiff order 4. Fixed
    timestepping only.


## Problems when not using IMEX:

### (ERROR: type ODEFunction has no field f1)
  - `SplitEuler`: 1st order fully explicit method. Used for testing accuracy
    of splits.
  - `IMEXEuler` : 1st order explicit Euler mixed with implicit Euler. Fixed time
    step only.
  - `CNAB2`: Crank-Nicolson Adams Bashforth Order 2. Fixed time step only.
  - `SBDF2` : 2nd order IMEX BDF method. Fixed time step only.
  - `SBDF3` : 3rd order IMEX BDF method. Fixed time step only. In development.
  - `SBDF4` : 4th order IMEX BDF method. Fixed time step only. In development.
  - `LawsonEuler` - First order exponential Euler scheme. Fixed timestepping only.
  - `NorsettEuler` - First order exponential-RK scheme. Fixed timestepping only. Alias: `ETD1`.
  - `ETD2` - Second order Exponential Time Differencing method (in development). Fixed timestepping only. Doesn't support Krylov approximation.
  - `ETDRK2` - 2nd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK3` - 3rd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK4` - 4th order exponential-RK scheme. Fixed timestepping only.
  - `HochOst4` - 4th order exponential-RK scheme with stiff order 4. Fixed
    timestepping only.
Basically all solvers are only defined on SplitODEProblems

### Not defined (ERROR: UndefVarError: `CNLF` not defined in `Main`)
  - `CNLF`: Crank-Nicolson Leapfrog of Order 2. Fixed time step only.


## Problems when using IMEX:

### ERROR: MethodError: no method matching size(::typeof(DispersiveShallowWater.rhs_split_stiff!))
The function `size` exists, but no method is defined for this combination of argument types.
(Maybe the problem here is how I implemented it but why do they want a size of a function.??)
  - `LawsonEuler` - First order exponential Euler scheme. Fixed timestepping only.
  - `NorsettEuler` - First order exponential-RK scheme. Fixed timestepping only. Alias: `ETD1`.
  - `ETD2` - Second order Exponential Time Differencing method (in development). Fixed timestepping only. Doesn't support Krylov approximation.
  - `ETDRK2` - 2nd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK3` - 3rd order exponential-RK scheme. Fixed timestepping only.
  - `ETDRK4` - 4th order exponential-RK scheme. Fixed timestepping only.
  - `HochOst4` - 4th order exponential-RK scheme with stiff order 4. Fixed
    timestepping only.

# Algorithms which work besides KenCarp:
  - `SplitEuler`: 1st order fully explicit method. Used for testing accuracy
    of splits.
  - `IMEXEuler` : 1st order explicit Euler mixed with implicit Euler. Fixed time
    step only.
  - `CNAB2`: Crank-Nicolson Adams Bashforth Order 2. Fixed time step only.
  - `CNLF`: Crank-Nicolson Leapfrog of Order 2. Fixed time step only.
  - `SBDF2` : 2nd order IMEX BDF method. Fixed time step only.
  - `SBDF3` : 3rd order IMEX BDF method. Fixed time step only. In development.
  - `SBDF4` : 4th order IMEX BDF method. Fixed time step only. In development.

But not sure a) how practical they are because they only support fixed time steps and b) I can not really compare them. (Besides saying `ImplicitEuler()` with no IMEX is slower then `IMEXEuler()` with IMEX)

cwittens:
> Thanks! So the only methods that worked are the KenCarp* algorithms? And they all share the same problematic behavior of worse performance with the IMEX formulation compared to a single RHS?

For the solitary wave initial condition ``KenCarp4()`` was faster with IMEX compared to a single rhs! (250ms compared to 210ms)

https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/207#discussion_r2106233549

Should I also test the over KenCarps on the manufactured solution and on the solitary wave or first focus on trying to use SciML operators, because this probably changes to results again if it works?

ranocha:
Thanks! SciMLOperators should have a significant influence, so working on that first would be nice.

cwittens:
> Thanks! SciMLOperators should have a significant influence, so working on that first would be nice.
## MatrixOperator
I tried using ``MatrixOperator`` and it significantly worsens performance / strange behaviour occurs. 

For now I only experimented locally with my own implementation first without using `ArrayPartition` (see later).

I define a ``MatrixOperator`` (e.g. to be able to use the exponential integrators)
```julia
opD3 = MatrixOperator(-1 / 6 * sqrt_gd * d^2 * D3)
```
and as comparisons
```julia
function rhs_stiff_op!(dη, η, p, t)
    (; opD3) = p
    mul!(dη, opD3, η)
    return nothing
end

function rhs_stiff!(dη, η, p, t)
    (; D3, g, d,) = p
    sqrt_gd = sqrt(g * d)
    mul!(dη, D3, η)
    @. dη .*= -1 / 6 * sqrt_gd * d^2 
    return nothing
end
```

Now I can define 
```julia
prob_split  = ODEProblem(SplitFunction(opD3,          rhs_nonstiff!), u0, tspan, p)
prob_split2 = ODEProblem(SplitFunction(rhs_stiff_op!, rhs_nonstiff!), u0, tspan, p)
prob_split3 = ODEProblem(SplitFunction(rhs_stiff!,    rhs_nonstiff!), u0, tspan, p)
```

And even thou now I can use  exponential integrators on `prob_split` (`prob_split2/3` gives an error as expected) they take of the order of tens of seconds (instead of ms) to run.  

And when using something like `KenCarp4`, where one is able to compare the results, I get: 
```julia
@btime sol_split  = solve(prob_split,  KenCarp4(), dt = 0.05, save_everystep=false); # => 1.353 s (7876 allocations: 19.22 MiB)
@btime sol_split2 = solve(prob_split2, KenCarp4(), dt = 0.05, save_everystep=false); # => 17.259 ms (251 allocations: 4.26 MiB)
@btime sol_split3 = solve(prob_split3, KenCarp4(), dt = 0.05, save_everystep=false); # => 16.673 ms (251 allocations: 4.26 MiB)
``` 
(where of cause one still needs to keep in mind that ``mul!(Vector, MatrixOperator, Vector)`` is pretty much as fast as ``mul!(Vector, SparseMatrix, Vector)``, which means considerably slower then using SBP Operators.)


## ArrayPartition
When doing the same as above but using ``ArrayPartition``, `sol_split` is not working anymore and gives the following error:
```julia
ERROR: MethodError: no method matching ArrayPartition{Float64, Tuple{Vector{Float64}}}(::UndefInitializer, ::Int64)
The type `ArrayPartition{Float64, Tuple{Vector{Float64}}}` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  (::Type{ArrayPartition{T, S}} where {T, S<:Tuple})(::Any)
   @ RecursiveArrayTools C:\Users\colli\.julia\packages\RecursiveArrayTools\iAeBW\src\array_partition.jl:27

Stacktrace:
  [1] Krylov.GmresSolver(m::Int64, n::Int64, memory::Int64, S::Type{ArrayPartition{Float64, Tuple{Vector{Float64}}}})
    @ Krylov C:\Users\colli\.julia\packages\Krylov\wgV9p\src\krylov_solvers.jl:2495
  [2] Krylov.GmresSolver(A::OrdinaryDiffEqDifferentiation.WOperator{…}, b::ArrayPartition{…}, memory::Int64)
    @ Krylov C:\Users\colli\.julia\packages\Krylov\wgV9p\src\krylov_solvers.jl:2513
  [3] init_cacheval(alg::KrylovJL{…}, A::OrdinaryDiffEqDifferentiation.WOperator{…}, b::ArrayPartition{…}, u::ArrayPartition{…}, Pl::LinearSolve.InvPreconditioner{…}, Pr::Diagonal{…}, maxiters::Int64, abstol::Float64, reltol::Float64, verbose::Bool, assumptions::OperatorAssumptions{…}; zeroinit::Bool)
    @ LinearSolve C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\iterative_wrappers.jl:223
  [4] solve!(cache::LinearSolve.LinearCache{…}, alg::KrylovJL{…}; kwargs::@Kwargs{…})
    @ LinearSolve C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\iterative_wrappers.jl:255
  [5] macro expansion
    @ C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\default.jl:339 [inlined]
  [6] solve!(::LinearSolve.LinearCache{…}, ::LinearSolve.DefaultLinearSolver; assump::OperatorAssumptions{…}, kwargs::@Kwargs{…})
    @ LinearSolve C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\default.jl:332
  [7] solve!
    @ C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\default.jl:332 [inlined]
  [8] #solve!#8
    @ C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\common.jl:307 [inlined]
  [9] solve!
    @ C:\Users\colli\.julia\packages\LinearSolve\KN05h\src\common.jl:306 [inlined]
 [10] #dolinsolve#2
    @ C:\Users\colli\.julia\packages\OrdinaryDiffEqDifferentiation\M5647\src\linsolve_utils.jl:29 [inlined]
 [11] dolinsolve
    @ C:\Users\colli\.julia\packages\OrdinaryDiffEqDifferentiation\M5647\src\linsolve_utils.jl:5 [inlined]
 [12] compute_step!(nlsolver::OrdinaryDiffEqNonlinearSolve.NLSolver{…}, integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, γW::Float64)
    @ OrdinaryDiffEqNonlinearSolve C:\Users\colli\.julia\packages\OrdinaryDiffEqNonlinearSolve\e4hoO\src\newton.jl:224
 [13] nlsolve!(nlsolver::OrdinaryDiffEqNonlinearSolve.NLSolver{…}, integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, cache::OrdinaryDiffEqSDIRK.KenCarp4Cache{…}, repeat_step::Bool)
    @ OrdinaryDiffEqNonlinearSolve C:\Users\colli\.julia\packages\OrdinaryDiffEqNonlinearSolve\e4hoO\src\nlsolve.jl:52
 [14] perform_step!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, cache::OrdinaryDiffEqSDIRK.KenCarp4Cache{…}, repeat_step::Bool)
    @ OrdinaryDiffEqSDIRK C:\Users\colli\.julia\packages\OrdinaryDiffEqSDIRK\YAA9M\src\kencarp_kvaerno_perform_step.jl:1009
 [15] perform_step!
    @ C:\Users\colli\.julia\packages\OrdinaryDiffEqSDIRK\YAA9M\src\kencarp_kvaerno_perform_step.jl:957 [inlined]
 [16] solve!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…})
    @ OrdinaryDiffEqCore C:\Users\colli\.julia\packages\OrdinaryDiffEqCore\bMOsj\src\solve.jl:620
 [17] __solve(::ODEProblem{…}, ::KenCarp4{…}; kwargs::@Kwargs{…})
    @ OrdinaryDiffEqCore C:\Users\colli\.julia\packages\OrdinaryDiffEqCore\bMOsj\src\solve.jl:7
 [18] __solve
    @ C:\Users\colli\.julia\packages\OrdinaryDiffEqCore\bMOsj\src\solve.jl:1 [inlined]
 [19] #solve_call#35
    @ C:\Users\colli\.julia\packages\DiffEqBase\zYZst\src\solve.jl:635 [inlined]
 [20] solve_call
    @ C:\Users\colli\.julia\packages\DiffEqBase\zYZst\src\solve.jl:592 [inlined]
 [21] #solve_up#44
    @ C:\Users\colli\.julia\packages\DiffEqBase\zYZst\src\solve.jl:1142 [inlined]
 [22] solve_up
    @ C:\Users\colli\.julia\packages\DiffEqBase\zYZst\src\solve.jl:1120 [inlined]
 [23] #solve#42
    @ C:\Users\colli\.julia\packages\DiffEqBase\zYZst\src\solve.jl:1057 [inlined]
 [24] var"##core#371"()
    @ Main C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:598
 [25] var"##sample#372"(::Tuple{}, __params::BenchmarkTools.Parameters)
    @ Main C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:607
 [26] _lineartrial(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; maxevals::Int64, kwargs::@Kwargs{})
    @ BenchmarkTools C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:186
 [27] _lineartrial(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters)
    @ BenchmarkTools C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:181
 [28] #invokelatest#2
    @ .\essentials.jl:1055 [inlined]
 [29] invokelatest
    @ .\essentials.jl:1052 [inlined]
 [30] #lineartrial#46
    @ C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:51 [inlined]
 [31] lineartrial
    @ C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:50 [inlined]
 [32] tune!(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; progressid::Nothing, nleaves::Float64, ndone::Float64, verbose::Bool, pad::String, kwargs::@Kwargs{})
    @ BenchmarkTools C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:299
 [33] tune!
    @ C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:288 [inlined]
 [34] tune!(b::BenchmarkTools.Benchmark)
    @ BenchmarkTools C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:288
 [35] top-level scope
    @ C:\Users\colli\.julia\packages\BenchmarkTools\1i1mY\src\execution.jl:728
Some type information was truncated. Use `show(err)` to see complete types.
```
Which to me looks like Krylov.GmresSolver just does not support ArrayPartitions.

Also interestingly (as also already stated above by Hendrik), when using the other solves also get quite a bit slower when using ArrayPartitions, even when "optimizing" the rhs! functions for ArrayPartitions, e.g. doing:
```julia
function rhs_stiff_op!(du::ArrayPartition{Float64, Tuple{Vector{Float64}}}, u::ArrayPartition{Float64, Tuple{Vector{Float64}}}, p, t)
    dη = du.x[1] # without unpacking it is even slower
    η = u.x[1]
    (; opD3) = p
    mul!(dη, opD3, η)
    return nothing
end
```
I get:
```julia
@btime sol_split  = solve(prob_split,  KenCarp4(), dt = 0.05, save_everystep=false); # => ERROR
@btime sol_split2 = solve(prob_split2, KenCarp4(), dt = 0.05, save_everystep=false); # => 352.997 ms (1263 allocations: 18.31 MiB)
@btime sol_split3 = solve(prob_split3, KenCarp4(), dt = 0.05, save_everystep=false); # => 406.871 ms (1263 allocations: 18.31 MiB)
```
Here I am not sure where all the allocations come from (sol_split.stats are identical)


## FunctionOperator
I still wasn't able to make FunctionOperators work, but in my defense, even following the ``getting_started`` tutorial is 1:1 gives errors (gonna write an issue).
So far I havent tried asking on the Discourse /Slack yet, because even when I can make it work without ArrayPartitions, I still feel like using them would not work for the same reason as MatrixOperator is not working.

cwittens:
On the topic of SciMLOperators and IMEX:
It looks like SciMLOperators are not fully implemented into OrdinaryDiffEq.jl, (since over two years) see: https://github.com/SciML/SciMLOperators.jl/issues/142

And it doesnt look like this will change soon. Here it was talked about being fixed in 1-2 weeks (as of May 2023)(https://github.com/SciML/SciMLOperators.jl/issues/191) and basically since then there have been no updates (https://github.com/SciML/OrdinaryDiffEq.jl/issues/2078#issuecomment-2229529314).

So with this knowledge in mind, I guess it wont make sense to try and make SciMLOperators work in DSW.jl (?)

ranocha:
Yes, it appears that this is indeed the case. @JoshuaLampert What's your take on this? Shall we just skip this part and focus on classical time integration methods and the remaining tasks?

JoshuaLampert:
It's a bit sad to see that the SciML ecosystem doesn't seem to be ready for IMEX methods for our use cases. I see roughly the following possibilities:

1. Ask the SciML people for help/ideas.
2. Finish this PR, set `split_ode = Val{false}()` as default, and merge it anyway since it adds the technical possibility to use IMEX methods, but doesn't destroy anything. The disadvantages of would be that it contradicts the principle to only add features that are really a benefit, users wanting to use IMEX methods, wonder why they are slow (we could document that properly), and that we need to maintain the additional code (should be fine, I think).
3. Just keep this open and continue to work on other things.

So, if you still have some motivation to continue the work on this @cwittens and would be willing to discuss this issue with the SciML folks, I would go with 1. But if not (which would be totally understandable), I tend to go with 3. for now and follow the status of SciMLOperators.jl and OrdinaryDiffEq.jl for any updates to eventually come back to this once there is a solution. That would of course a bit unfortunate because of the work you spend on this @cwittens. Thanks a lot anyway! At least, we learned something.

PR #208, opened 2025-05-23
added two kdv_1d to benchmarking tests

cwittens:
is this it?
and also ``tspan = (0.0, 1e-10)`` seems really small to me (?) Edit: its benchmarking the rhs! and not the solve, isnt it?


JoshuaLampert (dismissed):
Thanks! Yes, the example is only run for a very short time have access to the `semi`. It's then only benchmarking `rhs!`.

JoshuaLampert:
Maybe it would make more sense to benchmark the whole `solve` call because for the IMEX methods for the KdV equation, we do not call `rhs!`. Or is it possible to call the right hand side from the `ode`, which would be `rhs!` for the `ODEProblem` and the combination of stiff and non-stiff part for the `SplitODEProblem`, @ranocha?

ranocha:
Since we only used explicit methods so far, benchmarking the `rhs!` was fine. In particular, this reduced noise from the analysis callback etc. that we would get from benchmarking `solve`.

For implicit methods, it indeed makes more sense to benchmark the `solve` call since we may perform some tuning to help the implicit solvers.

cwittens:
> Since we only used explicit methods so far, benchmarking the `rhs!` was fine. In particular, this reduced noise from the analysis callback etc. that we would get from benchmarking `solve`.
> 
> For implicit methods, it indeed makes more sense to benchmark the `solve` call since we may perform some tuning to help the implicit solvers.

cant we then just also pass the argument ``callbacks = nothing`` to make sure we dont have the callback noise?

JoshuaLampert:
It's a bit weird why in [the new benchmark comment](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/208#issuecomment-2909778760) there are no entries for the new KdV examples. With the [old setup](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/208#issuecomment-2905262286) this used to work.

ranocha:
That's strange indeed. Maybe report this as an issue in their repo, merge this PR, and check whether it works fine for new PRs?

PR #209, opened 2025-05-23
Use `initial_condition_soliton` in more cases

ranocha:
This makes it easier to use the same (physically relevant) initial condition for both the hyperbolic approximation of SGN and the original equations. With the current release, we would have to use `initial_condition_soliton` for the hyperbolic version and `initial_condition_convergence_test` for the original equations.


JoshuaLampert (approved):
Thanks!

JoshuaLampert:
Copilot supports Julia now :smile:

PR #211, opened 2025-05-26
Update Benchmark action

JoshuaLampert:
AirspeedVelocity.jl now provides a simpler GitHub action that we can reuse. This should also post the memory in the comment.


ranocha:
Let's check?

ranocha (approved):

PR #214, opened 2025-06-03
Test benchmark action

JoshuaLampert:
Just to test that now the action runs the yaml file from the PR.


JoshuaLampert:
Seems to work now 👍

PR #216, opened 2025-06-07
test CI on Julia v1.12

JoshuaLampert:


ranocha:
```quote
JET.jl fails to precompile. I think JET.jl is only compatible with Julia v1.12 in newer versions than we currently allow (because newer versions are incompatible with Julia v1.10 and v1.11, cf. CompatHelper: bump compat for JET to 0.10 for package test, (keep existing compat) #194). So I it will probably be hard to let the tests run on both pre-1.12 and 1.12.
```

Would it make sense to move the JET.jl tests to a new CI job - or multiple of them - to fix this? Alternatively, it seems to be fine to use JET.jl only with a specific version, e.g., the LTS or the newest release.

JoshuaLampert:
> ```
> JET.jl fails to precompile. I think JET.jl is only compatible with Julia v1.12 in newer versions than we currently allow (because newer versions are incompatible with Julia v1.10 and v1.11, cf. CompatHelper: bump compat for JET to 0.10 for package test, (keep existing compat) #194). So I it will probably be hard to let the tests run on both pre-1.12 and 1.12.
> ```
> 
> Would it make sense to move the JET.jl tests to a new CI job - or multiple of them - to fix this? Alternatively, it seems to be fine to use JET.jl only with a specific version, e.g., the LTS or the newest release.

Would that work? We need to have some compat bound for JET.jl in test/Project.toml. If it is something like v0.9.9 the environment for Julia v1.12 cannot instantiate and if it is something like v0.10, the environment for Julia <v1.12 cannot instantiate even if we do not run any JET.jl tests in either of the tests. If I understand correctly, we would also need different environments for the different jobs or am I missing something?

JoshuaLampert:
@ranocha, do you have an idea how to fix the [world age issue](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/actions/runs/17202679280/job/48796324808?pr=216#step:7:28629) with `convergence_test`? Even with passing `@__MODULE__`  it fails.

ranocha:
> > ```
> > JET.jl fails to precompile. I think JET.jl is only compatible with Julia v1.12 in newer versions than we currently allow (because newer versions are incompatible with Julia v1.10 and v1.11, cf. CompatHelper: bump compat for JET to 0.10 for package test, (keep existing compat) #194). So I it will probably be hard to let the tests run on both pre-1.12 and 1.12.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would it make sense to move the JET.jl tests to a new CI job - or multiple of them - to fix this? Alternatively, it seems to be fine to use JET.jl only with a specific version, e.g., the LTS or the newest release.
> 
> Would that work? We need to have some compat bound for JET.jl in test/Project.toml. If it is something like v0.9.9 the environment for Julia v1.12 cannot instantiate and if it is something like v0.10, the environment for Julia <v1.12 cannot instantiate even if we do not run any JET.jl tests in either of the tests. If I understand correctly, we would also need different environments for the different jobs or am I missing something?

Yes - we could have a specific setup that installs JET.jl only for one of the Julia versions.

ranocha:
Does it help to use `Base.@invokelatest` at

https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/a8617f068f259dadc8acbea364606c6176d01fef/src/util.jl#L91

?

JoshuaLampert:
> Does it help to use `Base.@invokelatest` at
> 
> https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/a8617f068f259dadc8acbea364606c6176d01fef/src/util.jl#L91
> 
> (and does it change any performance significantly)?

Let's try in [14307b4](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/216/commits/14307b4a3c5c37db096f63ddafb36ee71eabcef9).

JoshuaLampert:
> > Does it help to use `Base.@invokelatest` at
> > https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/a8617f068f259dadc8acbea364606c6176d01fef/src/util.jl#L91
> > 
> > (and does it change any performance significantly)?
> 
> Let's try in [14307b4](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/216/commits/14307b4a3c5c37db096f63ddafb36ee71eabcef9).

```julia
l2_error, linf_error = @invokelatest mod.analysis_callback(mod.sol)
```
gives
```julia
ERROR: LoadError: UndefVarError: `sol` not defined in `Main.var"##625"`
The binding may be too new: running in world age 44347, while current world is 44365.
```
instead of
```julia
ERROR: LoadError: UndefVarError: `analysis_callback` not defined in `Main.var"##624"`
The binding may be too new: running in world age 44340, while current world is 44358.
```
Any idea how to solve this?

ranocha:
Did the plain `convergence_test` work as expected before when called from the REPL? Or does it also require the `invokelatest` fixes?

JoshuaLampert:
I tested it in Julia v1.11. So it worked. I didn't try from the REPL in Julia v1.12. But I can try.

JoshuaLampert (commented):
CI passes on v1.12 and pre-v1.12 :partying_face:
I would say, we can merge this once v1.12 is released, which shouldn't be in the too far future.
@ranocha, was the separate workflow for JET.jl roughly what you had in mind with

> Would it make sense to move the JET.jl tests to a new CI job - or multiple of them - to fix this?

.github/workflows/JET.yml
```diff
@@ -0,0 +1,86 @@
...
+jobs:
+  test:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '~1.12.0-0'
```
JoshuaLampert, line 51:
```suggestion
          - '1.12'
```

ranocha (commented):
Thanks, this looks good to me! We need to remmeber to add the new JET.jl CI job to the list of required tests for PRs.

PR #217, opened 2025-06-20
Added kdv_1d non dimensional solitary wave example

cwittens:
- added `prim2nondim` and `nondim2prim` 
- added example `kdv_1d_non_dim.jl` which solves the solitary wave solution for the non-dim KdV eq. by transforming it to the dimensional version, solving it in DSW.jl and after that converting the solution back to the non-dim case. Here I commeted out how I plot the final solution as it is common for the example files to end with sol, but as it is not straight forward, I wanted to include an example. (maybe there is also a better version as the one I currently use, which feels a bit like a hack)

Also even for large times ``tspan = (0.0, 150.0)`` the l2 error decreases to about 1e-9 (from ~1e-4) for N = 2048 and abs/reltol = 1e-11.


JoshuaLampert:
Regarding the plotting: I would have hoped we could just do `plot(semi => sol, conversion = prim2nondim)`. Would that also work and if not what is the problem?

JoshuaLampert:
Ah yes, but that makes sense. I think we should have a `varnames(::typeof(prim2nondim), ::KdVEquation1D{Float64})` anyway as we have with the other conversion functions [like here](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/e0aa9fd56c701a9b2aa4cb3f6df2618f6cee358a/src/equations/equations.jl#L86).

cwittens:
plot(semi => sol, conversion = prim2nondim, plot_initial = true) works now!
But I still need to add some tests for the new functions before this is ready.

cwittens:
I have added tests now besides for 
```julia
function prim2nondim(sol::ODESolution, equations::KdVEquation1D)
    sol_nondim = deepcopy(sol)  # Ensure we don't modify the original solution
    for i in eachindex(sol_nondim.u)
        eta = sol_nondim.u[i].x[1]  # Get the eta values (first and only variable for KdV)
        sol_nondim.u[i].x[1] .= prim2nondim.(eta, equations)  # Convert and update in-place
    end
    return sol_nondim
end
````
because here I am not sure if I should keep this function. 
Originally I made it to use `plot(semi => sol_nondim)`, but if one can just use `plot(semi => sol, conversion = prim2nondim)` this is not needed anymore. 
Also the functions _feels_ kinda different compared to other functions in DSW. 
One could argue it is useful but if one really wants to use DSW.jl to also calculate the non-dim version, one can probably just write something like this for oneself.

File: examples/kdv_1d/kdv_1d_non_dim.jl [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+using SummationByPartsOperators: upwind_operators, periodic_derivative_operator
+
+###############################################################################
+# Semidiscretization of the KdV equation 
+
+function initial_condition_non_dimensional(x, t, equations::KdVEquation1D, mesh)
+    c = 1 / 3
+    x_t = mod(x - c * t - xmin(mesh), xmax(mesh) - xmin(mesh)) + xmin(mesh)
+    u = 3 * c * sech(sqrt(9 * c) / 6 * x_t)^2
+    return u
```
JoshuaLampert, line 12:
```suggestion
    return SVector(u)
```
See below.

examples/kdv_1d/kdv_1d_non_dim.jl
```diff
@@ -0,0 +1,60 @@
...
+end
+
+function initial_condition_non_dimensional_converted(x, t, equations::KdVEquation1D, mesh)
+    u = initial_condition_non_dimensional(x, t, equations, mesh)
+    eta = nondim2prim(u, equations)
+    return SVector(eta)
```
JoshuaLampert, line 18: I would prefer calling `nondim2prim` on an `SVector` `u` instead of a scalar because in general these conversion functions usually assume an `SVector` input and also return an `SVector` output. I would like to keep this pattern also for `nondim2prim` and `prim2nondim`. I think it's fine how you define the functions below allowing both vectors and scalars, but I would still only call them on `SVector`s for consistency. See my suggestion above.
```suggestion
    return nondim2prim(u, equations)
```

src/equations/kdv_1d.jl
```diff
@@ -119,6 +119,139 @@ function source_terms_manufactured(q, x, t, equations::KdVEquation1D)
...
+
+See also [`nondim2prim`](@ref).
+"""
+function prim2nondim(eta, equations::KdVEquation1D)
+    u = @. eta / equations.D + 2 / 3
+    return u
+end
+
+"""
+    prim2nondim(sol::ODESolution, equations::KdVEquation1D)
```
JoshuaLampert, line 186: As you said, I would also vote for removing this function. We don't have such a functions for the other conversions like `prim2cons` or `prim2phys`. So I would also not define it for `prim2nondim`.

test/test_unit.jl
```diff
@@ -112,6 +113,13 @@ end
     @test @inferred(waterheight(q, equations)) == 43.0
     @test @inferred(still_water_surface(q, equations)) == 0.0
     @test @inferred(prim2phys(q, equations)) == @inferred(prim2prim(q, equations))
+
+    # Test non-dimensional conversion functions
+    eta = 42.0
```
JoshuaLampert, line 118:
```suggestion
```
I would simply reuse `q` from above.

test/test_unit.jl
```diff
@@ -112,6 +113,13 @@ end
...
+    equations_nondim = KdVEquation1D(gravity = 4 / 27, D = 3.0)
+    u_expected = eta / 3.0 + 2 / 3
+    @test @inferred(prim2nondim(eta, equations_nondim)) == u_expected
+    @test @inferred(nondim2prim(u_expected, equations_nondim)) == eta
```
JoshuaLampert, line 122:
```suggestion
    @test @inferred(nondim2prim(u_expected, equations_nondim)) == q
```

JoshuaLampert (changes requested):
Thanks! If I remember correctly, you also compared the result to a code, which directly uses the nondimensional formulation (either a self-defined solver or the one available at https://github.com/abhibsws/2024_kdvh_RR/blob/main/code.jl)? Could you post the difference between the two implementations, please?

JoshuaLampert:
For the Downgrade failure, could you try to bump [OrdinaryRosenbrock.jl to v1.5](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/e0aa9fd56c701a9b2aa4cb3f6df2618f6cee358a/test/Project.toml#L23). This should be the version, where they allowed LinearSolve.jl v3.

cwittens:
> For the Downgrade failure, could you try to bump [OrdinaryRosenbrock.jl to v1.5](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/e0aa9fd56c701a9b2aa4cb3f6df2618f6cee358a/test/Project.toml#L23). This should be the version, where they allowed LinearSolve.jl v3.

looks like this did not fix it. :(

JoshuaLampert:
> > For the Downgrade failure, could you try to bump [OrdinaryRosenbrock.jl to v1.5](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/e0aa9fd56c701a9b2aa4cb3f6df2618f6cee358a/test/Project.toml#L23). This should be the version, where they allowed LinearSolve.jl v3.
> 
> looks like this did not fix it. :(

Well, it looks like it fixed this particular one, but there is a next one🤦‍♂️

JoshuaLampert:
I can look it at once I am back at my computer.

cwittens:
> Thanks! If I remember correctly, you also compared the result to a code, which directly uses the nondimensional formulation (either a self-defined solver or the one available at https://github.com/abhibsws/2024_kdvh_RR/blob/main/code.jl)? Could you post the difference between the two implementations, please?

Before implementing KdV in DSW.jl I compared a self defined version of `rhs_kdv_dim` (which has the discretization I then implemented in DSW) and `rhs_kdv_nondim`. 
Now I only compared the dimensional solution I got from DSW.jl transformed back to be non-dimensional to the non-dimensional solitary wave solution.
Comparing the dimensional solution from DSW.jl transformed back to be non-dimensional to the numerical solution obtained by `rhs_kdv_nondim` does not seem to make to much sense.

But what I did not do and probably makes sense, is to compare it to non trivial initial conditions of the non-dim KdV eq (which then of course need to be solved by something like `rhs_kdv_nondim`).

cwittens:
> > Comparing the dimensional solution from DSW.jl transformed back to be non-dimensional to the numerical solution obtained by `rhs_kdv_nondim` does not seem to make to much sense.
> 
> Why? Wouldn't this be the same as what you describe at
> 
> > Before implementing KdV in DSW.jl I compared a self defined version of `rhs_kdv_dim` (which has the discretization I then implemented in DSW) and `rhs_kdv_nondim`.
> 
> just using DSW.jl instead of a self-defined version of the rhs? I would like to see that these two options give (almost) the same results. I think you showed me an image of the difference between the two solutions. I just wanted to make sure, we still get that result. Ok, we see that the errors to the analytical solution are small so it looks good, but I think having a direct comparison between the two methods, which should be theoretically equivalent (at least up to the semidiscrete level) would be nice.

My thinking was that if doing the transformation of the dimensionless initial condition to dimensions and the solving it using the dimensional rhs (in DSW.jl) and then transforming back to it being dimensionless, if this convergence to the dimensionless analytical solution, then of course it also converges to numerical solution using the dimensionless rhs, because the dimensionless rhs converges to the dimensionless analytical solution.

cwittens:
> Can you also a NEWS.md entry, please?

I added two possibilities. Just choose the one you like more

JoshuaLampert:
> My thinking was that if doing the transformation of the dimensionless initial condition to dimensions and the solving it using the dimensional rhs (in DSW.jl) and then transforming back to it being dimensionless, if this convergence to the dimensionless analytical solution, then of course it also converges to numerical solution using the dimensionless rhs, because the dimensionless rhs converges to the dimensionless analytical solution.

Yes, I agree. But the "of course" assumes that everything is correct (theory and implementation), which I am quite sure is the case, but I just wanted to double-check that by the numerical comparison of these two methods. Also just the fact that both versions converge to the analytical solution, it does not necessarily mean that both are computing the same thing. I would expect very small differences between both numerical solutions if both setups use very small tolerances for the time integration.

JoshuaLampert:
> > Can you also a NEWS.md entry, please?
> 
> I added two possibilities. Just choose the one you like more

Thanks! Let's take the shorter one because we usually have short entries. If someone is interested in the details, they can follow the link to the PR.

JoshuaLampert (approved):
Thanks!

ranocha (approved):
Thanks

PR #222, opened 2025-06-26
Fix initialization of hyperbolic approximation variable of SGN

### LINKED ISSUE #221: Initialization of the hyperbolic approximation parameters of SGN
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/b4cba0b29eb4569703ed96c5e2f53d5508432adb/src/equations/hyperbolic_serre_green_naghdi_1d.jl#L142-L144

For variable bathymetry, we should use 

$$w = \dot \eta + \frac{3}{2} \dot b = -h \nabla \cdot v + \frac{3}{2} v \cdot \nabla b$$

In the current implementation, the variable bathymetry part is missing.

JoshuaLampert:
Fixes #221.


src/equations/hyperbolic_serre_green_naghdi_1d.jl
```diff
@@ -139,9 +139,14 @@ function set_approximation_variables!(q, mesh,
     # H ≈ h
     @.. H = eta + D - equations.eta0 # h = (h + b) + (eta0 - b) - eta0
 
-    # w ≈ -h v_x
+    # w ≈ -h v_x + 3/2 v b_x
     mul!(w, D1, v)
     @.. w = -H * w
+    if !(equations.bathymetry_type isa BathymetryFlat)
+        b = bathymetry(q, equations)
+        b_x = D1 * b
```
JoshuaLampert, line 147: Unfortunately, we do not have access to `cache` here. So it's difficult to do the multiplication in-place. Do you see a way for an in-place multiplication or would you be fine with this version since it is also not so performance-critical (it's only computed once and not in `rhs!`), @ranocha?

ranocha (dismissed):
Thanks!

JoshuaLampert:
> Test values have to be updated...

See my comment [above](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/222#issuecomment-3009073170).
- Do you think the other tests do not fail because the changes are within the test tolerances?
- Can you provide the code you used for the computation of the manufactured solution or directly the new terms?

If these two points are clarified, I will update the test values.

ranocha (commented):
Looks reasonable to me. Do you know why the allocations increased?

JoshuaLampert:
> Looks reasonable to me. Do you know why the allocations increased?

Seems like it was spurious. After rerunning the tests, they are successful again.

ranocha (approved):
Thanks

PR #223, opened 2025-06-30
Reflecting boundary conditions for the SGN eqs.

cwittens:
This PR implements reflecting boundary conditions for the SGN eqs.

I made it in such a way that it reuses the periodic bc `rhs's`, by slightly changing the `assemble_system_matrix!` function (just whilewriting this down I am thinking if it makes sense to handle the reflecting bc case using multi dispatch instead of an if statement?) and uses the old `rhs's` by default when using reflecting bc and only when one has a Solver with a `var_coef_derivative_operator` and `bathymetry_flat` it will use 'old' reflecting bc `rhs` implementation.

The `var_coef_derivative_operator` was slightly faster thanthe 'normal' approach (~5%), but having it 'just work' with reflecting bc I think outweighs the cost (and one can of course still just use var_coef_derivative_operator).

Also, I implemented manufactured solutions for the BathymetryMildSlope/BathymetryVariable case. Here `h` and `v` are missing the exp() part, because this made the source terms loooooong. (*VSCode not responding, because the plain text was too long* - long)

About the `energy_total_modified!`, I am not sure if this is correct. It calculates it the same as for the periodic boundary condition case. As far as I see it, the distinction before about how to calculate `∫ 1/6 h^3 v_x^2` was just because one was able to use variable coefficient properties and not because the energy is different with reflecting bc, but as I said, I am not sure.

And here not related to the PR and maybe I am just late to the party, but I finally came around to write this down and it speeds up the *putting in the error values* - task quite a bit:
```julia
err1 = errors(analysis_callback)
int = integrals(analysis_callback)
clipboard("l2=$(err1.l2_error[:, end]),\nlinf=$(err1.linf_error[:, end]),\nchange_waterheight=$(int.waterheight_total[end] - int.waterheight_total[1]),\nchange_entropy_modified=$(int.entropy_modified[end] - int.entropy_modified[1])")
```


src/equations/serre_green_naghdi_1d.jl
```diff
@@ -383,14 +390,127 @@ function source_terms_manufactured_reflecting(q, x, t,
     return SVector(s1, s2, zero(s1))
 end
 
-# flat bathymetry with periodic boundary conditions
+"""
+    source_terms_manufactured_reflecting(q, x, t, equations::SerreGreenNaghdiEquations1D{BathymetryVariable}, mesh)
```
JoshuaLampert, line 394:
```suggestion
    source_terms_manufactured_reflecting(q, x, t,
                                                                           equations::Union{SerreGreenNaghdiEquations1D{BathymetryMildSlope},
                                                                                            SerreGreenNaghdiEquations1D{BathymetryVariable}})
```
Can you also remove the `mesh` argument in docstring of the other `source_terms_manufactured_reflecting` method above (and potentially other docstrings)?

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -383,14 +390,127 @@ function source_terms_manufactured_reflecting(q, x, t,
...
+    t4 = t^4
+
+    sinpix = sin(pi * x)
+    sin2pix = sin(2pi * x)
+    cospix = cos(pi * x)
+
+    sin2_pix = sin(pi * x)^2
+    cos2_pix = cos(pi * x)^2
+    cos3_pix = cos(pi * x)^3
+    cos4_pix = cos(pi * x)^4
```
JoshuaLampert, line 421:
```suggestion
    sinpix = sinpi(x)
    sin2pix = sinpi(2 * x)
    cospix = cospi(x)

    sin2_pix = sinpi(x)^2
    cos2_pix = cospi(x)^2
    cos3_pix = cospi(x)^3
    cos4_pix = cospi(x)^4
```

src/equations/serre_green_naghdi_1d.jl
```diff
@@ -1239,11 +1375,13 @@ function energy_total_modified!(e, q_global,
         scale_by_inverse_mass_matrix!(e, D1)
 
         @.. e = e + 1 / 2 * g * eta^2 + 1 / 2 * h * v^2
+
```
JoshuaLampert, line 1378:
```suggestion
```

JoshuaLampert (changes requested):
Thanks a lot! Only a first brief review. I agree it would be nice to handle the reflecting bc case using multiple dispatch instead of an if statement. Then it would be easier to extend from outside.
I think, we have to reconsider the implementation for `energy_total_modified!`. IIRC, the version for periodic boundary conditions relied on integration by parts, where boundary terms were neglected (because of the periodic bc). This means, we cannot use the same formulation here as for the periodic boundary conditions, but rather something more like the variable coefficients case. But I would need to look into this more closely.

src/equations/serre_green_naghdi_1d.jl
JoshuaLampert: I mean everywhere, where there is a parameter `mesh` in the docstring, but the function does not get a parameter `mesh`. The initital conditions get the `mesh`, so we should keep it there.

JoshuaLampert (changes requested):
Thanks! Looking at it again, I think your implementation of `energy_total_modified!` works.
1. Could you also provide a `convergence_test` result for the manufactured solution with variable bathymetry?
2. What does the following give?
```julia
julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_reflecting.jl", abstol = 1.0e-14, reltol = 1.0e-14); plot(semi => sol, plot_initial = true)
[...]

julia> integrals(analysis_callback).entropy_modified |> extrema |> x -> last(x) - first(x)
```
3. Did you try using relaxation?

Do I understand correctly that

4. before, only a variable coefficient operator was supported for a flat bathymetry and reflecting boundary conditions and now we also support central and upwind operators for this case?
5. this is not breaking because the variable coefficient case can be done in the same way as before?
6. the previous two examples with reflecting bc didn't change, but the files were only renamed to include the "_var_coef_op" suffix?
7. all 3 * 2 * 2 = 12 combinations of `bathymetry_flat`, `bathymatry_mild_slope`, and `bathymetry_variable` with `boundary_conditions_periodic` and `boundary_conditions_reflecting` solved by central, and upwind operators are supported and a variable coefficient operator is only supported for `boundary_conditions_reflecting` and `bathymetry_flat`? If yes, is there a fundamental issue with the combination variable coefficient operator, `boundary_conditions_reflecting` and a non-flat bathymetry?

ranocha:
> If yes, is there a fundamental issue with the combination variable coefficient operator, `boundary_conditions_reflecting` and a non-flat bathymetry?

I just do not see how to make it work in a nice way - but maybe I am just too stupid/blind. The issue is that the additional terms that appear, e.g., in the elliptic operator, also include first-order derivatives in the presence of a variable bathymetry. I do not see how to use the variable-coefficient operator in a nice way there to still be able to prove ellipticity (positive definiteness of the matrix) and get energy conservation.

JoshuaLampert:
> > If yes, is there a fundamental issue with the combination variable coefficient operator, `boundary_conditions_reflecting` and a non-flat bathymetry?
> 
> I just do not see how to make it work in a nice way - but maybe I am just too stupid/blind. The issue is that the additional terms that appear, e.g., in the elliptic operator, also include first-order derivatives in the presence of a variable bathymetry. I do not see how to use the variable-coefficient operator in a nice way there to still be able to prove ellipticity (positive definiteness of the matrix) and get energy conservation.

Ah, I see. Yes, this doesn't seem completely straightforward. In this case, it would be fine for me to only support variable coefficient operators for flat bathymetry (and, of course, only for reflecting boundary conditions) as long as an understandable error message is thrown when one tries to use a variable coefficient operator with variable bathymetry.

cwittens:
I decided to add to `Semidiscretization`
```julia
    if (isa(solver.D1, AbstractNonperiodicDerivativeOperator) ||
        isa(solver.D2, AbstractNonperiodicDerivativeOperator) ||
        isa(solver.D3, AbstractNonperiodicDerivativeOperator)) &&
       isa(boundary_conditions, BoundaryConditionPeriodic)
        throw(ArgumentError("Non-periodic derivative operators in `solver` are incompatible with periodic boundary conditions."))
    end
```
to catch all kinds of errors relating to this early (also for different equations) instead of something like
```julia
    if D1 isa DerivativeOperator && D2 isa VarCoefDerivativeOperator
        if boundary_conditions isa BoundaryConditionPeriodic
            throw(ArgumentError("Periodic boundary conditions need periodic operators"))
        end
       [...]
```

Also I added a warning to `create_cache` for variable bathymetry. Thy way it is currently handled is that if one calls SGN with a Solver with VarCoefDerivativeOperator and non-flat bathymetry and reflecting boundary conditions, it will just ignore `solver.D2` and just calculate with `rhs_sgn_central!`. But one could of cause also throw an error here.
```julia
    if D1 isa DerivativeOperator && D2 isa VarCoefDerivativeOperator
        @warn "VarCoefDerivativeOperators are currently only supported for reflecting boundary conditions with flat bathymetry. DispersiveShallowWater.jl will ignore `solver.D2` and just use `solver.D1` in this case."
    end
```

JoshuaLampert:
> I just locally added `relaxation_callback = RelaxationCallback(invariant = entropy_modified)` to serre_green_naghdi_soliton_reflecting.jl and it does not throw an error. But also `integrals(analysis_callback).entropy_modified |> extrema |> x -> last(x) - first(x)` gives the same value with or without `relaxation_callback`.

I was surprised by that. So I also checked that locally and I got (note that I didn't pass any tolerances, but I also get something of this order with passing the tolerances):

```julia
julia> trixi_include("examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_reflecting.jl");
[...]
julia> integrals(analysis_callback).entropy_modified |> extrema |> x -> last(x) - first(x)
5.684341886080802e-14
```

So the relaxation seems to work. Did you double-check that you passed the `relaxation_callback`as the first argument to the `CallbackSet`, see, e.g., https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/ae1b1ab8d028fdead8e91c350b74e12233aedeff/examples/bbm_bbm_1d/bbm_bbm_1d_upwind_relaxation.jl#L44.

cwittens:
> So the relaxation seems to work. Did you double-check that you passed the `relaxation_callback`as the first argument to the `CallbackSet`, see, e.g.,

Not sure if this was the problem or that after trying without the `relaxation_callback`, I did not execute 
```julia
analysis_callback = AnalysisCallback(semi; interval = 100,
                                     extra_analysis_errors = (:conservation_error,),
                                     extra_analysis_integrals = (waterheight_total,
                                                                 entropy_modified))
```
again, so that `integrals(analysis_callback).entropy_modified` had both the old and the new values inside of it. 

Not sure if this is on purpose, but maybe (if it is easy to do) `analysis_callback` should get re-instantiated at every `solve` call.

But now, doing it the correct way, I also get `5.684341886080802e-14`.

JoshuaLampert:
> Not sure if this is on purpose, but maybe (if it is easy to do) `analysis_callback` should get re-instantiated at every `solve` call.

I see. We could maybe set `tstops`, `errors`, and `integrals` of the `analysis_callback` back to empty vectors in the `initialize!` function: 
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/ae1b1ab8d028fdead8e91c350b74e12233aedeff/src/callbacks_step/analysis.jl#L152

cwittens:
> I see. We could maybe set `tstops`, `errors`, and `integrals` of the `analysis_callback` back to empty vectors in the `initialize!` function:

[9d346e3](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/223/commits/9d346e31282f276f4ce621b0ade8425cde15e9d6) should do it.

JoshuaLampert:
> > I see. We could maybe set `tstops`, `errors`, and `integrals` of the `analysis_callback` back to empty vectors in the `initialize!` function:
> 
> [9d346e3](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/223/commits/9d346e31282f276f4ce621b0ade8425cde15e9d6) should do it.

Great, thanks! I'll try to review this PR later today.

JoshuaLampert (approved):
Thanks! We could consider adding an example using the `RelaxationCallback` or add it only to the tests, but I would also be fine without.

JoshuaLampert (changes requested):
Could you also add a NEWS.md entry, please?

NEWS.md
```diff
@@ -10,8 +10,9 @@ for human readability.
 
 #### Added
 
+
```
JoshuaLampert, line 13:
```suggestion
```
We usually only have one newline here.

cwittens: I started adding a new line before I realized it makes more sense to just add to the existing one and apparently messed up 🫣

JoshuaLampert (approved):
Thanks!

PR #226, opened 2025-07-24
Use more functions from TrixiTest.jl

JoshuaLampert:
With TrixiTest.jl v0.1.3 we have some more functionality in the common TrixiTest.jl, which can be reused here. Note that we need a custom `@test_trixi_include_base` because it looks quite different compared to the one in TrixiTest.jl, but I also adapted to the new style, where we have a `@test_trixi_include_base` and a `@test_trixi_include` to stay consistent and avoid confusion.
Regarding the `@test_allocations`, we now also need to pass the `rhs!` to the macro.

<details>
<summary>For potential future reference, here is the script I used to replace all occurrences</summary>

```julia
for (root, _, files) in walkdir("test")
    for file in files
        if endswith(file, ".jl")
            filename = joinpath(root, file)
            (tmppath, tmpio) = mktemp()
            open(filename) do io
                for line in readlines(io)
                    if contains(line, "@test_allocations(semi, sol, ")
                        new_line = replace(line, "@test_allocations(semi, sol, " => "@test_allocations(DispersiveShallowWater.rhs!, semi, sol, ")
                        println(tmpio, new_line)
                    else
                        println(tmpio, line)
                    end
                end
            end
            close(tmpio)
            mv(tmppath, filename, force = true)
        end
    end
end

```

</details>


ranocha (approved):
Thanks!

PR #227, opened 2025-07-26
Use julia-downgrade-compat@2

JoshuaLampert:
With the new julia-downgrade-compat v2, the action switched to use [Resolver.jl](https://github.com/StefanKarpinski/Resolver.jl) to find minimal compatible versions. If everything works as intended, this makes it MUCH simpler to work with and debug the downgrade action because in case of incompatible lower bounds, you don't need to resolve the incompatibilities manually by finding out a set of working minimal bounds. This should now be done by Resolver.jl.


JoshuaLampert:
Hm, the Downgrade job seems to just use the same versions as the normal job (see https://github.com/NumericalMathematics/DispersiveShallowWater.jl/actions/runs/16541187956/job/46782631560?pr=227#step:8:57 and https://github.com/NumericalMathematics/DispersiveShallowWater.jl/actions/runs/16541187950/job/46782631572?pr=227#step:7:52) meaning it doesn't really test the compat bounds. Same also happens in https://github.com/SciML/SciMLBase.jl/pull/1075. I think the new version is not quite ready yet.

ranocha (dismissed):
Sounds like a good plan. Thanks!

JoshuaLampert:
This should finally be ready.

ranocha (approved):
Thanks for all the work you put into fixing and updating downgrade actions!

PR #229, opened 2025-07-27
Fix bug when plotting with conversion function

JoshuaLampert:
There was a bug when plotting the solution with a conversion function, which has more variables than the default `prim2phys`, where one frame was left empty.

```julia
julia> include("examples/hyperbolic_serre_green_naghdi_1d/hyperbolic_serre_green_naghdi_soliton.jl");
[...]
julia> plot(semi => sol, plot_initial = true, legend = :bottomleft, conversion = prim2prim)
```

gave before this PR:
before
and after this PR gives:
after

Something similar also happened with `prim2cons` for instance.

### DIFF
--- a/src/visualization.jl
+++ b/src/visualization.jl
@@ -55,13 +55,15 @@ end
     plot_title --> "$(get_name(equations)) at t = $(round(t, digits=5))"
     layout --> nsubplots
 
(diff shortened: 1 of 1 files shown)


cwittens (approved):
LGTM!

PR #230, opened 2025-07-27
Add dispersion relation for `KdVEquation1D`

JoshuaLampert:
I noticed, we forgot to add the linear dispersion relation for the `KdVEquation1D`. However, something seems to be not quite right. Running the [example in the docs](https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/dispersion/) for `equations = kdv` gives me
traveling_waves_kdv

For some reason, the numerical solution of the `KdVEquation1D` doesn't have the correct phase velocity. It is way too small (does is even move at all?). The speed of the green dot moves with the speed from the linear dispersion relation, which looks reasonable as it is of almost the same speed as the blue dot. The green wave should have the same speed (or in other words: The green dot should always stay on the maximum of the wave). I don't see why the numerical solution doesn't get the velocity right. Do you have any idea, @cwittens?

### DIFF
--- a/docs/src/dispersion.md
+++ b/docs/src/dispersion.md
@@ -46,6 +46,10 @@ euler = EulerEquations1D(; gravity = g, eta0 = eta0)
 c_euler = wave_speed.(disp_rel, euler, k; normalize = true)
 plot(k, c_euler, label = "Euler", xlabel = "k", ylabel = "c / c_0", legend = :topright)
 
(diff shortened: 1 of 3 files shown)


JoshuaLampert:
Thanks to @cwittens there was an error in the dispersion relation, which is fixed now. With this fix the expected velocity is indeed almost zero. For smaller wave numbers the phase speed of the KdV equations matches the one from the Euler equation much closer and the corresponding gif show the same behavior. For example, for k = 1:
traveling_waves_kdv

PR #232, opened 2025-07-29
added error warning to KdV

cwittens:
not sure if I should also have a test for this third error case but I just realised if D3 == nothing KdV will make strange stuff


JoshuaLampert:
That's a good idea. What about the other equations? What happens if we do not pass a second-order operator? Maybe it makes sense to check this in general and also already during the creation of the `Semidiscretization`. We could add a `check_solver(equations, solver)` (or similar) function called within `Semidiscretization` that dispatches on `equations` and throws an error if an operator is not defined, which is needed.

cwittens:
I currently dont have the time to do 

> That's a good idea. What about the other equations? What happens if we do not pass a second-order operator? Maybe it makes sense to check this in general and also already during the creation of the `Semidiscretization`. We could add a `check_solver(equations, solver)` (or similar) function called within `Semidiscretization` that dispatches on `equations` and throws an error if an operator is not defined, which is needed.

Should I close this PR and open an issue so that it is not forgotten?

JoshuaLampert:
Ok, no worries. I would also be fine with keeping this PR open. I think then we also won't forget about it :)

src/semidiscretization.jl
```diff
@@ -88,6 +89,8 @@ function Semidiscretization(mesh, equations, initial_condition, solver;
                                                       solver, cache)
 end
 
+check_solver(equations, solver, boundary_conditions) = nothing
```
cwittens, line 92:
```suggestion
"""
    check_solver(equations, solver, boundary_conditions)

Check that the `solver` is compatible with the given `equations` and 
`boundary_conditions`. The default implementation performs no checks.
Specific equation types can override this method to validate that 
required derivative operators are present (e.g., some equations 
require `D2` or `D3` to be non-`nothing`).

Throws an `ArgumentError` if the solver is incompatible.
"""
check_solver(equations, solver, boundary_conditions) = nothing
```

cwittens (commented):
LGTM.
Does this need a News.md entry or not because it is internal only?

JoshuaLampert (approved):

ranocha:
I guess this can be merged, @JoshuaLampert?

PR #235, opened 2025-08-21
Add first JOSS paper draft

JoshuaLampert:
**DO NOT MERGE THIS PR**

We can use this PR to check and discuss the paper draft. The PR can be closed when the JOSS paper is finished. The branch `paper-2025-joss` must be kept around forever.

You can download the PDF of the paper draft from the GitHub action `Build JOSS paper draft PDF / JOSS Paper Draft (pull_request)` below.
- Click on it
- Go to `Summary` on the left
- Download the artifact

PR #236, opened 2025-08-22
Update Zenodo title

ranocha:
We have to use the same title for the paper developed in #235 and Zenodo.


.zenodo.json
```diff
@@ -1,7 +1,7 @@
 {
     "description": "Structure-Preserving Numerical Methods for Dispersive Shallow Water Models",
```
JoshuaLampert, line 2:
```suggestion
    "description": "A Julia library of structure-preserving numerical methods for dispersive wave equations",
```
Should we maybe make this consistent?

JoshuaLampert (commented):
Thanks!

PR #238, opened 2025-08-26
Add comparison with experimental data to docs

JoshuaLampert:


JoshuaLampert:
@ranocha, what would you suggest where to put the experimental data? In the main repo like now, in a gist, which is downloaded, or something else?

docs/src/dingemans.md
```diff
@@ -169,32 +169,89 @@ legend_plot = plot(legend=:top, framestyle=:none, legendfontsize=11)
...
 
 The results show how different dispersive wave models capture the wave evolution over the trapezoidal bathymetry.
 
+## Comparison with Experimental Data
+
+During the experiment of Dingemans, the wave evolution was recorded at six gauges along the flume. These measurements provide a reference for validating the numerical models.
+We compare the numerical results with the experimental data at the corresponding gauge locations.
+
+```@example dingemans
+path_dingemans = joinpath(@__DIR__, "data", "Dingemans", "Dingemans.csv")
```
ranocha, line 198: It will be easier to keep the data in this repository. However, it may be better to make them easily accessible to people using DispersiveShallowWater.jl (without having to work with the `docs` directory explicitly. What do you think about adding a function that returns the data and refers to Dingemans works in the docstring?

Such a function may read a CSV file - or just store the data in the function itself. I am not sure which approach will be better.

src/util.jl
```diff
@@ -47,6 +47,19 @@ function default_example()
     joinpath(examples_dir(), "bbm_bbm_1d", "bbm_bbm_1d_basic.jl")
 end
 
+"""
+    data_dir()
+
+Return the directory where the data files provided with DispersiveShallowWater.jl are located. If DispersiveShallowWater is
```
ranocha, line 53:
```suggestion
Return the directory where the data files provided with DispersiveShallowWater.jl are located. If DispersiveShallowWater.jl is
```

(since you used `.jl` when mentioning the package for the first time in this docstring).

ranocha:
Why are the results for the BBM-BBM equations so bad? The plot in https://numericalmathematics.github.io/DispersiveShallowWater.jl/previews/PR238/dingemans/#Visualization-of-Temporal-Evolution does not look good for BBM-BBM. If I remember correctly, we had reasonable results for BBM-BBM in your paper, didn't we?

ranocha:
Thanks! This looks better.

> If you agree, I would keep it like this and add a sentence why we use upwind operators for BBM-BBM and Svärd-Kalisch and central for the others. 

👍 

> Should we use upwind or central for SGN? 

I would expect upwind to perform at least as good as or better than the central one, so I would check the upwind version.

> If we want to use an upwind operator, should it also have `accuracy_order - 1` (like [here](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/7b422e9467e0c5544895c104db1ad4c820a0c3e9/examples/serre_green_naghdi_1d/serre_green_naghdi_soliton_upwind.jl#L26))? Why was that needed again for SGN?

If I remember correctly, the upwind operators are only used for higher-order derivatives. In particular, they will be used symmetrically for the second-derivative terms. Thus, these terms will have an even order of accuracy, i.e., the accuracy will be one order higher than that of the individual upwind operators. The central first-derivative operator induced by the upwind operators has the same property, i.e., setting `accuracy_order = 5` for the upwind operators results in a sixth-order accurate scheme in the end. Thus, I would typically choose the fifth-order upwind operators since they have a smaller stencil and will be faster than the sixth-order operators.

We can briefly comment on this in the docs and choose `accuracy_order = 5` for all upwind discretizations.

ranocha (dismissed):

ranocha (approved):

PR #244, opened 2025-10-22
Mention that models are one-dimensional

JoshuaLampert:
We should clarify that DispersiveShallowWater.jl only implements one-dimensional models (up to now). Suggested in https://github.com/openjournals/joss-reviews/issues/8894#issuecomment-3433515211.


jatkinson1000:
Sigh, I forgot to paste in the link.
It's late here...

Here you go: https://github.com/Cambridge-ICCS/green-ci

PR #245, opened 2025-10-23
Do not run benchmarks on non-code changes

ranocha:
See https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/244#issuecomment-3433915355

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@
   <img height="200px" alt="logo" src="https://github.com/user-attachments/assets/d2bccfbd-19c9-4448-a5d9-6ac927c32cc5">
 </p>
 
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -2,6 +2,19 @@ name: Benchmark this PR
 on:
   pull_request_target:
     branches: [ main ]  # change to your default branch
+    paths-ignore:
+      - 'CITATION.bib'
+      - 'LICENSE.md'
+      - 'README.md'
+      - '.zenodo.json'
+      - '.github/workflows/benchmark.yml'
```
JoshuaLampert:
```suggestion
```
I think usually we want to run benchmarks if we change the benchmark action file itself.

JoshuaLampert (approved):
Thanks!

PR #248, opened 2025-11-03
Update to JET.jl v0.11

JoshuaLampert:
See https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/246#issuecomment-3477953392.
Trying this locally gave a lot of errors from JET.jl, but they looked real. So let's see what CI says.

### DIFF
--- a/src/equations/serre_green_naghdi_1d.jl
+++ b/src/equations/serre_green_naghdi_1d.jl
@@ -1019,9 +1019,6 @@ function rhs_sgn_central!(dq, q, t, equations, source_terms, solver, cache,
         # Compute all derivatives required below
         (; h, h_x, v_x, h_hpb_x, b, b_x, hv_x, v2_x,
         h2_v_vx_x, h_vx_x, p_h, p_x, tmp) = cache
-        if equations.bathymetry_type isa BathymetryVariable
-            (; psi) = cache
-        end
 
         @.. b = equations.eta0 - D
         @.. h = eta - b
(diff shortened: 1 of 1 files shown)


ranocha:
> That's a bit tricky. I see why JET.jl complains about [this part](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/43798eeff64bca798a32d389dd2f134d2829e6e3/src/equations/serre_green_naghdi_1d.jl#L1022-L1064). But it's not so easy to fix it nicely because we cannot merge the if statements easily and it's also not so nice to always unpack psi again in each if statement.

You can report this as a sub-optimal result of JET.jl. From looking at the code, I do not see an obvious way how `psi` could not be defined there.

To fix this right now, the only option I see is to unpack `psi` every time from the `cache`.

JoshuaLampert:
> > That's a bit tricky. I see why JET.jl complains about [this part](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/43798eeff64bca798a32d389dd2f134d2829e6e3/src/equations/serre_green_naghdi_1d.jl#L1022-L1064). But it's not so easy to fix it nicely because we cannot merge the if statements easily and it's also not so nice to always unpack psi again in each if statement.
> 
> You can report this as a sub-optimal result of JET.jl. From looking at the code, I do not see an obvious way how `psi` could not be defined there.

Yes, JET.jl is not smart enough to realize that the first first `if equations.bathymetry_type isa BathymetryVariable` condition will always be true and therefore `psi` will be defined when the other `if equations.bathymetry_type isa BathymetryVariable` conditions are true. That's also basically the same issue with the first problem from `@trixi_timeit` in TrixiBase.jl.
I now always unpack `psi` from the `cache`.

ranocha:
Did you file an issue in their repo? They should know about false positives like this.

JoshuaLampert:
> Did you file an issue in their repo? They should know about false positives like this.

No because this is the same as https://github.com/aviatesk/JET.jl/issues/749. So the compiler cannot infer that in the second branch the variable is defined and we need to resort to workarounds. I am not a big fan of adding assertions in the branches as was also suggested in the issue above. So I think the only way to make JET.jl pass is using the workarounds we already found (unpacking `psi` in every branch and code duplication in TrixiBase.jl).

PR #251, opened 2025-11-07
Fix rendering in docs

JoshuaLampert:
I noticed that some parts were rendering incorrectly in https://numericalmathematics.github.io/DispersiveShallowWater.jl/stable/plotting/#Error-Analysis-Visualization. This should fix it.


ranocha (approved):
Thanks!

PR #262, opened 2025-11-18
Clarify parameters of Svärd-Kalisch equations

### LINKED ISSUE #258: Svaerd-Kalisch model
The implemented Svaerd-Kalisch model differs from the model in the published paper (JCP - equations 11-12). This model has 3 parameters, and when α or γ are non-zero, it is not suitable for modeling water wave propagation. This can be observed by noting that the phase velocity is not an even function of the wave number. As a side note, the convention for computing this quantity for negative wave numbers is not consistent across all models (Euler and Svaerd-Kalisch models are opposite to all other models). We suggest adding a note in the documentation to clearly indicate the parameter restrictions and warn users about the configurations that are not appropriate.

```julia
using DispersiveShallowWater
using Plots

eta0 = 0.0
reference_height() = 0.8
h0 = reference_height()
g = 9.81
disp_rel = LinearDispersionRelation(h0)
k = -5.0:0.01:5.0

bbm = BBMEquation1D(; gravity = g, eta0 = eta0, D = h0)
c_bbm = wave_speed.(disp_rel, bbm, k; normalize = true)
plot(k, c_bbm, label = "BBM")

sk = SvaerdKalischEquations1D(; gravity = g, eta0 = eta0,
                              alpha = 0.0, beta = 0.2308939393939394, gamma = 0.04034343434343434)
c_sk = wave_speed.(disp_rel, sk, k; normalize = true)
plot!(k, abs.(c_sk), label = "Svärd-Kalisch")
```

[JOSS review #9361](https://github.com/openjournals/joss-reviews/issues/9361)

JoshuaLampert:
Does this resolve your issue in #258, @pnavaro?
Closes #258.

### DIFF
--- a/src/equations/svaerd_kalisch_1d.jl
+++ b/src/equations/svaerd_kalisch_1d.jl
@@ -25,6 +25,12 @@ The gravitational acceleration `gravity` is denoted by ``g`` and the bottom topo
 The water height above the bathymetry is therefore given by
 ``h = \eta - \eta_0 + D``.
 
+!!! warning "Choice of parameters"
+    Note that the above equations are the more general form of the equations presented in the [preprint](https://arxiv.org/abs/2302.09924) by Svärd and Kalisch (2023)
+    including ``\alpha`` and ``\gamma`` terms. The [published version](https://doi.org/10.1016/j.jcp.2024.113516) considers only the case ``\alpha = \gamma = 0``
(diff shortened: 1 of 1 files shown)


pnavaro (approved):

ranocha (approved):
Thanks!

PR #275, opened 2025-12-15
Add reference to JOSS paper

JoshuaLampert:


ranocha (approved):
Thanks! Please create a new release so that the new info is contained in the stable docs.

PR #278, opened 2025-12-19
Always use return in formatter

JoshuaLampert:
This is the analogous PR as https://github.com/trixi-framework/Trixi.jl/pull/2707. Here, we can enforce the return from JuliaFormatter.jl though because we do not use `@kernel`s from KernelAbstractions.jl.


ranocha (approved):
Thanks!

PR #288, opened 2026-03-25
Implement hyperbolized Sainte-Marie equations developed by Escalante et al.

ranocha:
~I will add a few additional modifications here soon; I just opened this PR now to get the PR number for the NEWS.md entry.~

I have implemented the hyperbolization of the Sainte-Marie equations proposed by [Escalante, Dumbser, and Castro (2019)](https://doi.org/10.1016/j.jcp.2019.05.035). The energy-conserving semidiscretization is adapted from the conservative version derived in https://arxiv.org/abs/2603.18978.

Convergence test with manufactured solution:

```julia
julia> convergence_test("examples/hyperbolic_sainte_marie_1d/hyperbolic_sainte_marie_manufactured.jl", 4);
[...]
l2
η                        v                        D                        w                        p                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   8.79e-03  -         50   5.67e-03  -         50   0.00e+00  -         50   6.08e-02  -         50   2.17e-01  -         
100  5.43e-04  4.02      100  3.44e-04  4.04      100  0.00e+00  NaN       100  3.71e-03  4.04      100  1.37e-02  3.99      
200  3.32e-05  4.03      200  2.15e-05  4.00      200  0.00e+00  NaN       200  2.32e-04  4.00      200  8.54e-04  4.00      
400  2.07e-06  4.01      400  1.38e-06  3.97      400  0.00e+00  NaN       400  1.45e-05  4.00      400  5.34e-05  4.00      

mean           4.02      mean           4.00      mean           NaN       mean           4.01      mean           4.00      
linf
η                        v                        D                        w                        p                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   2.49e-02  -         50   1.02e-02  -         50   0.00e+00  -         50   1.32e-01  -         50   3.97e-01  -         
100  1.80e-03  3.79      100  6.42e-04  4.00      100  0.00e+00  NaN       100  7.60e-03  4.12      100  2.24e-02  4.15      
200  1.19e-04  3.92      200  4.06e-05  3.99      200  0.00e+00  NaN       200  4.55e-04  4.06      200  1.49e-03  3.92      
400  7.72e-06  3.94      400  2.55e-06  3.99      400  0.00e+00  NaN       400  2.83e-05  4.01      400  9.30e-05  4.00      

mean           3.88      mean           3.99      mean           NaN       mean           4.06      mean           4.02      

```

CC @MarcoArtiano


JoshuaLampert (changes requested):
Thanks a lot!
Before I take a closer review, could you please add the new equation system to the README, the landing page of the docs, the table in https://numericalmathematics.github.io/DispersiveShallowWater.jl/previews/PR288/overview/#eq_overview, and the docstring in the overview page https://numericalmathematics.github.io/DispersiveShallowWater.jl/previews/PR288/overview/#Detailed-Documentation?
Could you also please add some unit tests [here](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/679edf86f933313a9a65bb97130417c1baf047d9/test/test_unit.jl) similar to the other equations and increase coverage?
I also have two questions:
- Is there a known linear dispersion relation for the (hyperbolic) Saint-Marie equations? If yes, could you please add it to https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/679edf86f933313a9a65bb97130417c1baf047d9/src/dispersion_relation.jl and if no, could you track that in https://github.com/NumericalMathematics/DispersiveShallowWater.jl/issues/131, please?
- Out of interest: Did you do any comparison with your implementation in https://github.com/MarcoArtiano/2026_nonconservative_sbp using Trixi.jl? Would it maybe even be possible to create a simple setup, where both implementations should compute the same (e.g., central flux in the Trixi.jl version vs discontinuosly coupled SBP operators here or using `FDSBP` with one element and a global FD SBP operator in the Trixi.jl version vs the same global SBP operator here)? If not or if it is too much work, that's also fine. I'm just curious.

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,428 @@
...
+  [DOI: 10.5802/smai-jcm.66](https://doi.org/10.5802/smai-jcm.66)
+- Escalante, Dumbser and Castro (2019)
+  An efficient hyperbolic relaxation system for dispersive non-hydrostatic
+  water waves and its solution with high order discontinuous Galerkin schemes
+  [DOI: 10.1016/j.jcp.2019.05.035](https://doi.org/10.1016/j.jcp.2019.05.035)
+
+The energy-conserving semidiscretization is basically taken from the
+following reference, but adapted to the standard style of
+DispersiveShallowWater.jl to use the primitive variables `q = (η, v, D, w, p)`.
+The semidiscretization implemented here is derived from their method with parameters ``\alpha_1 = 2``, ``\alpha_2 = 0``, ``\alpha_3 = 0``, and ``\alpha_4 = 0``.
```
MarcoArtiano, line 73: Out of curiosity, is there a particular reason for choosing these coefficients?

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,428 @@
...
+    alpha::RealT # hyperbolic relaxation parameter (→ ∞ for Sainte-Marie)
+    c_squared::RealT # c^2 = alpha^2 * g * eta0
+end
+
+function HyperbolicSainteMarieEquations1D(; bathymetry_type = bathymetry_mild_slope,
+                                          gravity,
+                                          eta0 = 0.0,
+                                          h0,
+                                          alpha = 3.0)
+    c_squared = alpha^2 * gravity * h0
```
MarcoArtiano, line 94:
```suggestion
if alpha <= 1
    error("Invalid alpha: must be > 1 to satisfy model assumptions. Got alpha = $alpha.")
    end
    c_squared = alpha^2 * gravity * h0
```

Should we add a check about the value of $$\alpha$$?

ranocha:
Thanks for the initial review! I updated the docs and added unit tests.

- I added the new equations to #131, tracking missing dispersion relations.
- I added the new equations to #95, tracking missing functionality (reflecting BCs).

MarcoArtiano (changes requested):
Thanks! I've left a few comments,

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: These coefficients agree with what Mario Richiutto and I used for the hyperbolized Serre-Green-Naghdi equations ($$\alpha_1$$) and make a strong-form implementation of the nonconservative terms easier (other parameters).

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: I thought about this as well. However, choosing `alpha = 0` together with initial data `w = p = 0` makes the system equivalent to the standard shallow-water equations, which is useful for testing. Thus, I decided not to implement such a check.

ranocha:
I added the dispersion relations.

ranocha:
> Out of interest: Did you do any comparison with your implementation in https://github.com/MarcoArtiano/2026_nonconservative_sbp using Trixi.jl? Would it maybe even be possible to create a simple setup, where both implementations should compute the same (e.g., central flux in the Trixi.jl version vs discontinuosly coupled SBP operators here or using FDSBP with one element and a global FD SBP operator in the Trixi.jl version vs the same global SBP operator here)? If not or if it is too much work, that's also fine. I'm just curious.

No, I didn't. I just derived the corresponding numerical method by hand. However, I would like to have the equations in TrixiShallowWater.jl. Maybe that's something @MarcoArtiano could do based on his implementation in the reproducibility repository? I spoke with Andrew: it would be totally fine to have a first version without all the wetting and drying treatment (which would also require additional wave breaking models for real applications).

docs/src/overview.md
```diff
@@ -26,10 +27,11 @@ The following table provides an overview of all available equation systems and t
 - ``D``: Still-water depth
 - ``w``: Auxiliary variable in hyperbolic approximation (``\approx -h v_x``)
```
JoshuaLampert: This is not quite true for Sainte-Marie because there $w$ approximates $w \approx -h v_x / 2 + v b_x$, right?

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,429 @@
...
+  [DOI: 10.48550/arXiv.2603.18978](https://arxiv.org/abs/2603.18978)
+"""
+struct HyperbolicSainteMarieEquations1D{Bathymetry <:
+                                        Union{BathymetryFlat, BathymetryMildSlope},
+                                        RealT <: Real} <: AbstractSainteMarieEquations{1, 5}
+    bathymetry_type::Bathymetry # type of bathymetry
+    gravity::RealT # gravitational acceleration
+    eta0::RealT # constant still-water surface
+    alpha::RealT # hyperbolic relaxation parameter (→ ∞ for Sainte-Marie)
+    c_squared::RealT # c^2 = alpha^2 * g * eta0
```
JoshuaLampert: The classical $\eta_0$ vs $h_0$ question: Here you use `eta0`, but below in the computation of `c_squared` you use `h0`. I think it should be `h0`, right?

```suggestion
    c_squared::RealT # c^2 = alpha^2 * g * h0
```

examples/hyperbolic_sainte_marie_1d/hyperbolic_sainte_marie_conservation.jl
```diff
@@ -0,0 +1,92 @@
...
+
+using OrdinaryDiffEqLowStorageRK
+using DispersiveShallowWater
+
+#=
+You can run this example manually and check the conservation properties as follows:
+
+julia> using Revise; using DispersiveShallowWater, OrdinaryDiffEqVerner
+
+julia> trixi_include("examples/hyperbolic_sainte_marie_1d/hyperbolic_sainte_marie_conservation.jl", tol = 1.0e-12, alg = Vern9());
```
JoshuaLampert, line 17: Just for my understanding: You use a high-accuracy time integrator with very low tolerance in this comment to avoid loss of conservation of the modified entropy due to time integration? Alternatively, we could also use the `RelaxationCallback` to guarantee conservation with other time integrators?

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,429 @@
...
+    initial_condition_manufactured(x, t, equations::HyperbolicSainteMarieEquations1D, mesh)
+
+A smooth manufactured solution in combination with
+[`source_terms_manufactured`](@ref).
+"""
+function initial_condition_manufactured(x, t,
+                                        equations::HyperbolicSainteMarieEquations1D,
+                                        mesh)
+    eta = 2 + cospi(2 * (x - 2 * t))
+    b = -5 - 2 * cospi(2 * x)
```
JoshuaLampert: This also sets `b` like this if we pass `bathymetry_flat`. Should we catch this case by setting `b` to zero similar to https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/679edf86f933313a9a65bb97130417c1baf047d9/src/equations/serre_green_naghdi_1d.jl#L132-L136?

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,429 @@
...
+    s2 = pi * (-a2 + a5 * (4 * a6 + 2 * a4) / a14 + 2 * g * a4 + a11 + 2 * a9)
+    s3 = zero(s1)
+    s4 = -2 * a5 / a14 +
+         pi_2 * (14 * a8^2 - 4 * a2 * (a6 - a4) + a8 * (a14 + 12 * a12 + 2 * a13))
+    s5 = -pi * (3 + 2 * a8) * a9
+
+    return SVector(s1, s2, s3, s4, s5)
+end
+
+dingemans_calibration(equations::HyperbolicSainteMarieEquations1D) = 1.0
```
JoshuaLampert, line 248: Looking at the zoomed plot for t = 14.0 at https://numericalmathematics.github.io/DispersiveShallowWater.jl/previews/PR288/dingemans/, it looks like there is some phase shift, which might be reduced by tweaking this calibration parameter. How did you come up with 1.0 so far?

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,429 @@
...
+        # Plain: h_t + (h v)_x = 0
+        #
+        # Split form for energy conservation:
+        # h_t + h_x v + h v_x = 0
+        @.. dh = -(h_x * v + h * v_x)
+        if boundary_conditions isa BoundaryConditionReflecting
+            # dh[1] -= h[1] * v[1] / left_boundary_weight(D1)
+            # dh[end] += h[end] * v[end] / right_boundary_weight(D1)
+            # FIXME
+            error("Reflecting boundary conditions are not implemented yet.")
```
JoshuaLampert:
```suggestion
            throw(ArgumentError("Reflecting boundary conditions are not implemented yet."))
```
I think we usually want to avoid using plain `error` because it does not cause the simulation to fail. On the other hand, currently it is impossible to enter this if branch because of the dispatching. So we might also just keep it as it is until reflecting boundary conditions are implemented. Do you plan to add reflecting boundary conditions?

src/dispersion_relation.jl
```diff
@@ -128,3 +128,32 @@ function (disp_rel::LinearDispersionRelation)(equations::SerreGreenNaghdiEquatio
...
+
+# See, e.g., non-numbered equation below eq. (14) in
+# - Escalante, Dumbser and Castro (2019)
+#   An efficient hyperbolic relaxation system for dispersive non-hydrostatic
+#   water waves and its solution with high order discontinuous Galerkin schemes
+#   [DOI: 10.1016/j.jcp.2019.05.035](https://doi.org/10.1016/j.jcp.2019.05.035)
+function (disp_rel::LinearDispersionRelation)(equations::SainteMarieEquations1D, k)
+    h0 = disp_rel.ref_height
+    g = gravity(equations)
+    return sqrt(g * h0) * k / sqrt(1 + (k * h0)^2 / 4)
```
JoshuaLampert, line 142:
```suggestion
    return sqrt(g * h0) * k * sqrt(1 + (k * h0)^2 / 4)
```
Maybe I'm missing something, but the equation below eq. (14) looks to me like the `sqrt(1 + (k * h0)^2 / 4)` term should be in the enumerator instead of in the denominator.

JoshuaLampert (changes requested):
Thanks! This looks already quite good. I left some comments below.

docs/src/overview.md
ranocha:
```suggestion
- ``w``: Auxiliary variable in hyperbolic approximation (e.g., ``\approx -h v_x`` for the hyperbolic SGN equations)
```

?

examples/hyperbolic_sainte_marie_1d/hyperbolic_sainte_marie_conservation.jl
ranocha: Yes, exactly. However, relaxation may also enforce conservation of an invariant if the spatial semidiscretization does not conserve it perfectly (depending on the spatial and temporal error). Thus, using strict tolerances helps me to check properties of the spatial discretization.

JoshuaLampert: Yes, makes sense. Thanks!

src/dispersion_relation.jl
ranocha: I think this is a typo in their paper. If you compare eq. (14) to the next one, there should be a division. Moreover, (15) also needs a division to get the same result when $$\alpha \to \infty$$.

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: Do you think we really need an additional manufactured solution for flat bathymetry? I would have to derive the corresponding source terms for that case.

src/dispersion_relation.jl
JoshuaLampert: Yes, you are right. The equation below eq. (14) should be $\frac{C_J^2}{gH} = J(kH)^{-1}$.

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Alternatively, we could restrict `initial_condition_manufactured` and `source_terms_manufactured` to `HyperbolicSainteMarieEquations1D{BathymetryMildSlope}`. Otherwise one might want to run a simulation with `initial_condition_manufactured` and `source_terms_manufactured` together with flat bathymetry and wonders why the solution does not converge.

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: I tried setting the value so that the time series at the first few gauges look reasonable. Note that in the zoomed plots, the phase shifts at t = 14 and t = 70 are in opposite directions. Thus, we cannot fix both of them together.

If you have another suggestion for a good calibration parameter (based on how you estimated it for other models), please let me know.

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Probably one would run into https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/679edf86f933313a9a65bb97130417c1baf047d9/src/semidiscretization.jl#L242 though. I would still prefer to restrict the dispatching.

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Hm, true, do you have an explanation of why the hyperbolic Sainte-Marie equations have such a big difference in the phase velocity compared to the other equations? It seems like the waves are too fast. The linear dispersion relation of them does not seem so bad. We could try a setup similar to the lower one in https://numericalmathematics.github.io/DispersiveShallowWater.jl/previews/PR288/dispersion/ to see if the numerical solution matches the velocity predicted by the linear dispersion relation.

docs/src/dispersion.md
JoshuaLampert, line 90:
```suggestion
equations = hsm
```
See the other comment. For a wave number of $k = 3$ from the plot of the linear dispersion relations, it looks like the wave speed of the HSM equations should almost exactly match the one of the Euler equations (slightly too slow maybe). Let's see.

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Ok, just quickly tried that in [1f06885](https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/288/commits/1f06885939a1a59953dfce1f9a4b8560549f1f4d) and it looks as expected. It's just slightly slower.

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: I implemented a manufactured solution for flat bathymetry:

```julia
julia> convergence_test("examples/hyperbolic_sainte_marie_1d/hyperbolic_sainte_marie_manufactured.jl", 4, bathymetry_type = bathymetry_flat);
[...]
l2
η                        v                        D                        w                        p                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   8.91e-03  -         50   2.82e-03  -         50   0.00e+00  -         50   1.87e-02  -         50   2.47e-01  -         
100  5.52e-04  4.01      100  1.76e-04  4.00      100  0.00e+00  NaN       100  1.16e-03  4.00      100  1.55e-02  3.99      
200  3.45e-05  4.00      200  1.11e-05  4.00      200  0.00e+00  NaN       200  7.27e-05  4.00      200  9.73e-04  4.00      
400  2.16e-06  4.00      400  7.16e-07  3.95      400  0.00e+00  NaN       400  4.54e-06  4.00      400  6.09e-05  4.00      

mean           4.00      mean           3.98      mean           NaN       mean           4.00      mean           4.00      
linf
η                        v                        D                        w                        p                        
N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       N    error     EOC       
50   2.32e-02  -         50   3.58e-03  -         50   0.00e+00  -         50   5.03e-02  -         50   4.00e-01  -         
100  1.46e-03  3.99      100  2.25e-04  3.99      100  0.00e+00  NaN       100  3.19e-03  3.98      100  2.42e-02  4.05      
200  9.14e-05  4.00      200  1.41e-05  4.00      200  0.00e+00  NaN       200  2.00e-04  4.00      200  1.54e-03  3.97      
400  5.74e-06  3.99      400  1.38e-06  3.35      400  0.00e+00  NaN       400  1.25e-05  4.00      400  9.61e-05  4.00      

mean           3.99      mean           3.78      mean           NaN       mean           3.99      mean           4.01      
```

docs/src/dispersion.md
```diff
@@ -87,7 +87,7 @@ solve the problem with the Svärd-Kalisch equations.
 using OrdinaryDiffEqTsit5
 using Printf
 
-equations = sk
+equations = hsm
```
ranocha: This should be reverted before merging this PR, shouldn't it?

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Cool, thanks!

docs/src/dispersion.md
JoshuaLampert: Yes. It was just a quick test.

ranocha:
```suggestion
equations = sk
```

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: Yes, that seems to be the reason. Note that the dispersion relation for $$\alpha = 3$$ is also quite a bit different from the limiting Sainte-Marie system.

ranocha: If you cannot find a bug in my code, it looks like this is just how the hyperbolized Sainte-Marie system behaves 🤷

JoshuaLampert: > Note that the dispersion relation for α = 3 is also quite a bit different from the limiting Sainte-Marie system.

Yes, but for the Dingemans experiment we are talking about a wave number of around [0.85](
https://github.com/NumericalMathematics/DispersiveShallowWater.jl/blob/679edf86f933313a9a65bb97130417c1baf047d9/src/equations/equations.jl#L642) (that probably becomes bigger when the waves become steeper, but it should not become too high) and for small wave numbers both linear dispersion relations look very similar and even better than most of the other models we have implemented. So I'm wondering where this shift comes from :thinking:
But yes, if we don't find any bug, we can just go on and live with it.

src/dispersion_relation.jl
```diff
@@ -128,3 +128,32 @@ function (disp_rel::LinearDispersionRelation)(equations::SerreGreenNaghdiEquatio
     g = gravity(equations)
     return sqrt(g * h0) * k / sqrt(1.0 + (k * h0)^2 / 3)
 end
+
+# See, e.g., non-numbered equation below eq. (14) in
+# - Escalante, Dumbser and Castro (2019)
+#   An efficient hyperbolic relaxation system for dispersive non-hydrostatic
+#   water waves and its solution with high order discontinuous Galerkin schemes
+#   [DOI: 10.1016/j.jcp.2019.05.035](https://doi.org/10.1016/j.jcp.2019.05.035)
```
JoshuaLampert, line 136:
```suggestion
#   [DOI: 10.1016/j.jcp.2019.05.035](https://doi.org/10.1016/j.jcp.2019.05.035)
# There is a typo in the paper: The equation below eq. (14) should be ``\frac{C_J^2}{gH} = J(kH)^{-1}``
# to be consistent with eq. (14) and eq. (15).
```
Such that we don't forget that there is a typo in the reference paper?

JoshuaLampert (changes requested):
Thanks! If @MarcoArtiano agrees, this can be merged once my final small comment below is resolved.

src/equations/hyperbolic_sainte_marie_1d.jl
```diff
@@ -0,0 +1,464 @@
...
+total water height ``\eta = h + b`` and the velocity ``v``.
+The gravitational acceleration `gravity` is denoted by ``g`` and the bottom topography
+(bathymetry) ``b = \eta_0 - D``. The water height above the bathymetry
+is therefore given by ``h = \eta - \eta_0 + D``.
+The total water height is therefore given by ``\eta = h + b``.
+
+There are two additional variables compared to the [`SainteMarieEquations1D`](@ref):
+the auxiliary variable ``w \approx -h v_x / 2 + v b_x`` and the non-hydrostatic pressure ``p``.
+In the formal limit ``c^2 \to \infty``, the hyperbolic approximation recovers the original Sainte-Marie system.
+Escalante, Dumbser and Castro (2019) choose the hyperbolization parameter as ``c = \alpha \sqrt{g h_0}`` for some background water height ``h_0``.
```
MarcoArtiano, line 32:
```suggestion
Escalante, Dumbser and Castro (2019) choose the hyperbolization parameter as ``c = \alpha \sqrt{g h_0}`` for some background water height ``h_0`` and ``\alpha > 1``.
```

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: I would be happy to find such a bug - but I haven't been able to do so so far...

MarcoArtiano (changes requested):
LGTM! I'm just wondering whether we should somewhere write that `\alpha` should indeed be greater than 1.

I will open a PR in TrixiShallowWater.jl for the conservative form implementation.

src/equations/hyperbolic_sainte_marie_1d.jl
MarcoArtiano: Thanks!

src/equations/hyperbolic_sainte_marie_1d.jl
ranocha: Good idea. I just pushed a similar (but more detailed) comment in 01f0e5a.

docs/src/dingemans.md
```diff
@@ -116,12 +122,13 @@ ode_bbmbbm = semidiscretize(semi_bbmbbm, tspan)
 ode_sk = semidiscretize(semi_sk, tspan)
 ode_sgn = semidiscretize(semi_sgn, tspan)
 ode_hysgn = semidiscretize(semi_hysgn, tspan)
-
+ode_hysm = semidiscretize(semi_hysm, tspan)
```
JoshuaLampert, line 125:
```suggestion
ode_hysm = semidiscretize(semi_hysm, tspan)

```

src/equations/hyperbolic_sainte_marie_1d.jl
JoshuaLampert: Should we also add the information that $\alpha = 0$ together with $p = w = 0$ recovers the SWE?

ranocha: I just added such a note in 0d29bd75d226a39774a8b851e409ab29a07481d3

JoshuaLampert: Thanks!

JoshuaLampert (approved):
Thanks a lot for adding the new equation system, @ranocha and thanks @MarcoArtiano for your review! From my side, this can be merged.

src/equations/hyperbolic_sainte_marie_1d.jl
MarcoArtiano: Thanks!

MarcoArtiano:
> Thanks a lot for adding the new equation system, @ranocha and thanks @MarcoArtiano for you review! From my side, this can be merged.

Thank you too!

PR #304, opened 2026-05-01
replace CompatHelper by Dependabot

ranocha:
This setup will group all Julia package updates into a single PR. This will be more annoying, e.g., if we cannot update one dependency but some others. However, this situation should occur less often than the typical case we had with CompatHelper.


JoshuaLampert (approved):
Thanks!

> This setup will group all Julia package updates into a single PR. This will be more annoying, e.g., if we cannot update one dependency but some others. However, this situation should occur less often than the typical case we had with CompatHelper.

Yes, I agree. Most of the time if more than one package updates the major version within one day (or if it happens on the weekend within a few days because `interval: "daily"` means only weekdays, see [here](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#interval)), they belong together and should be considered together (like the recent SciML updates). In this case, grouping is helpful.

